PR #28, opened 2014-12-30
Externalize Injection sources

naokikimura:


h3xstream:
Initially, I was really thinking it would be a bad idea.
Here are few reasons mainly around the tests
- A important practice of the project is to test every rules/signatures. Quickly adding a line of configuration without confirming that the rule is working as expected is a bit against this objective.
- Currently, the test coverage confirm that the code is working as expected not just placebo tests. 1 line code become 1 line of configuration but the coverage metric would be loss.

The more I think about it, I think their might be gain to create a __injection detector separated__ from the default (jdo, jpa, etc). Custom method signatures would be in configuration files but the known signature would stay in the code.
This would serve the purpose of adding quickly a signature while doing a code review.

What do you think about it ?

naokikimura:
I agree with your thoughts.

It is very important to test the signature. At the same time, it is also important that more people to extend it is easy to use.

In the case of well-known APIs (for example Apache Commons DBUtils) you should add a signatures with the test as ever.

However, it is not possible to include many of the API to code of this plug-in.

So, there is a very significance that you can dynamically add a signature by using a configuration file through the system properties.

I think that dynamically added signatures should be tested by who did.

h3xstream:
For the reason mention previously. I will accept the change if it does not externalize the current detectors of injection but for additional signatures only.

PR #30, opened 2015-01-13
Copy all files in metadata folder

jsotuyod:
- This allows plugins to specify other files such as bugrank.txt and adjustBugrank.txt

Without this, all detectors added by the tested plugin which have a base rank above 10 will be ignored, since the BugRank computation will always double the default score, making it impossible to test those with this library (which means everything except `SECURITY` and `CORRECTNESS`). You probably aren't seeing this, as all your plugin's category are `SECURITY`, which has a [rank of 5](https://code.google.com/p/findbugs/source/browse/findbugs/etc/bugrank.txt); which doubled is 10, and is still under the default threshold of 20.

If you changed the category of any pattern to anything else, such as `BAD_PRACTICE` you would notice the detector is never run, as it's filtered by the `ExecutionPlan`


h3xstream:
Cool. Good to know.

PR #36, opened 2015-01-24
Enable additional signatures to detector of injection

naokikimura:
I was fixed Pull #28.

### DIFF
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/injection/AdditionalInjectionSource.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/injection/AdditionalInjectionSource.java
@@ -1,0 +1,292 @@
+/**
+ * Find Security Bugs
+ * Copyright (c) Philippe Arteau, All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3.0 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library.
+ */
+package com.h3xstream.findsecbugs.injection;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


h3xstream:
Quick comment.. I saw your pull request when you submitted. I was busy with other dev.
I will include the change it in 1.3.2. Before, I need to publish few mijor bug fixes with version 1.3.1.

Stay tune.

h3xstream:
@naokikimura 
I have a question. How are you adding new method signatures ?
Are you repackaging the plugin or you are creating a dummy plugin that is classpath accessible ?

naokikimura:
There are two ways to add signatures.

One is to specify the signature definition files with the `com.h3xstream.findsecbugs.injection.sql.sources.files` system property.

```
findbugs -jvmArgs "-Dcom.h3xstream.findsecbugs.injection.sql.sources.files=${PWD}/InjectionSource.properties" \
 -textui -pluginList findsecbugs-plugin-1.3.2-SNAPSHOT.jar ./classes/
```

The other is to put the signature definition file on this place `custom-definitions/com/h3xstream/findsecbugs/injection/sql/InjectionSource.properties`. it's the same place of the plugin. If the place of plugin is `./plugin/findsecbugs-plugin-1.3.2-SNAPSHOT.jar`, you need to put the signature definition file on `./plugin/custom-definitions/com/h3xstream/findsecbugs/injection/sql/InjectionSource.properties`.

h3xstream:
I came to the conclusion that it will probably not be usable with most tools (Maven, Sonar, etc). The command line and the IDE will make it usable.

I will do small change to the description before the next release by rephasing the description to be more "generic" and by adding some directive on how to add new signature.

Again thanks for your contribution and patience. (I was really hesitant with this PR.)

naokikimura:
Thank you for merging it.

> I came to the conclusion that it will probably not be usable with most tools (Maven, Sonar, etc).

This feature in the Sonar can not be used perhaps, but in the Maven can be used. Setting of signature definition files uses the `jvmArgs` parameters of findbugs-maven-plugin. For example, you can write the pom.xml as follows:

    <project>
      ...
      <reporting>
        <plugins>
            <plugin>
                <groupId>org.codehaus.mojo</groupId>
                <artifactId>findbugs-maven-plugin</artifactId>
                <version>2.5.2</version>
                <configuration>
                    <jvmArgs>-Dcom.h3xstream.findsecbugs.injection.sql.sources.files=${basedir}/src/findsecbugs/InjectionSource.properties</jvmArgs>
                    <plugins>
                        <plugin>
                            <groupId>com.h3xstream.findsecbugs</groupId>
                            <artifactId>findsecbugs-plugin</artifactId>
                            <version>1.3.2-SNAPSHOT</version>
                        </plugin>
                    </plugins>
                </configuration>
            </plugin>
        </plugins>
      </reporting>
      ...
    </project>

h3xstream:
@naokikimura Ah! Perfect.

PR #46, opened 2015-04-24
Detector for hard coded passwords and cryptographic keys

formanek:
Detection of hard coded password in FindBugs works only for database passwords and detectors in FindSecurityBugs are very specific too. I have made quite complex detector for many different sinks having passwords or keys as input parameters of type String, char[], byte[] or BigInteger. Since I am not using real program flow analysis (for now), false positives are possible but quite rare (I have tested it using some existing code).

BTW: I had to made some changes (but I did not commit them) to be able to build the project - changing a line number in a test of XSS detection in JSP page to pass, copying of licence file to roots of modules and adding header to source code of 5 files from your last commit.


h3xstream:
This is looks perfect!
I like the idea to look for the field name of constant value and correlate their load operation.

Thanks @formanek for sharing those signatures !

h3xstream:
I will do more extensive tests and review later on.

PR #49, opened 2015-04-29
Update messages_ja.xml

naokikimura:
I have translated the added messages to messages.xml.


h3xstream:
This is excellent ! I was just about to ask you for the android message.

PR #53, opened 2015-05-04
ECB and no integrity detection + tests

formanek:
I think bugs should be reported for default cipher instances (without specified mode) too. I have also written a test for the whole class and refactor it a bit.


h3xstream:
Good point on the default mode being ECB.
Excellent contribution!

PR #60, opened 2015-06-11
Allow analyze to set classpath entries

mbmihura:
- This allows to test scenarios where bugs occur due to interactions with
    code from third-party libraries.
 - Overrides BaseDetector.analyze() and FindbugsLauncher.analyze() to allow
    adding paths to auxClassPath during tests execution.
 - Add utility methods to get paths to jars in src/test/resources.


h3xstream:
Thanks for the contribution on obscure internal utility classes!

I really like it.


>This allows to test scenarios where bugs occur due to interactions with code from third-party libraries.

I am curious. Does it means that I could do signatures to identify vulnerable versions of libraries being used?

Issue #14, opened 2013-12-13
SQL Injection detected when StringBuilder is used with constant strings


domak:
Hi, 

As soon as I see a StringBuilder in JPA EntityManager.createQuery(), an SQL injection is detected:

```
StringBuilder buff = new StringBuilder("from Tiers t");
// add conditional where clauses

TypedQuery<Tiers> query1 = em.createQuery(buff.toString(), Tiers.class); // SQL injection detected
query1.getResultList();
```

Is it possible to improve detector by checking that no input parameter (method parameter or attribute) has been added to the StringBuilder? 

On the code I try to verify, there are a lot of where clauses added if a parameter is not null (and the developper uses setParameter() method)

h3xstream:
The initial code sample with StringBuilder is now supported. Even much more advanced flow analysis are done in **1.4.2**.
The release version will be available very soon.

ThrawnCA:
I'm still getting false positives from SQL_INJECTION_HIBERNATE when assembling a constant string:

    private static final String NTP_QUERY = new StringBuilder(
                    " select os.name, p.dateProcessed, cd.title || ' ' || cd.givenName || ' ' || cd.familyName, p.amount ")
            .append("   from Statement as st ")
            .append("  inner join st.paymentRequest as pr ")
            .append("  inner join pr.onlineService as os ")
            .append("   left outer join pr.contactDetails as cd ")
            .append("  inner join st.payment as p ")
            .append("  where p.dateProcessed >= :startDate ")
            .append("    and p.dateProcessed <= :endDate")
            .append("  order by p.id")
            .toString();

and then passing it to Hibernate Session `createQuery`.

h3xstream:
@ThrawnCA Please open a new ticket for this case.
Field are different beast.

Unless the type is primitive `final` instance can be altered. Here it is an immutable type `String` with a twist. In general, we didn't need to worry about constant String because the compiler does optimization were it will copy the content of constant everywhere they are used (inline into the code of methods). This late initialization does not allow the compiler to optimize the use of the string.

I think it is worth to be supported.

h3xstream:
I have created #385

PR #63, opened 2015-07-01
Improved hardcoded passwords and key detector

formanek:
New version of the detector with higher accuracy. The most important change is that constant values are tracked to a given argument of password or key sink method using flow analysis instead of only guessing the link between constant and method. Heuristics is used to catch positives with too complicated flow, but in contrast to previous detector, fields with suspicious names are reported all for a class in one report and only if they were not reported by high-confidence analysis. More information such as hard coded source is added to the bug report (if available) too. I have also integrated OAuth detector into this one (just by altering configuration).


h3xstream:
Quick update : The pull request will be integrated soon. I just did not get the time to look at the changes set.

formanek:
I have not created a new branch for taint analysis so commits are here... I have implemented flow analysis, which is tracking "taintness" (taint, uknown, safe or null) at a given code location. Currently, state "taint" is never set (I am going to add it later for explicitly configured taint sources such as reading from file), but constant strings or objects are marked safe. There is a configuration for "transfer methods", where can be specified, how parameters transfer their "taintness" to the return value, to other parameters or to the object invoking the configured method. TaintDetector uses the information from TaintDataFlow to set the priority of the reported bug. All test passed (with some priority changes) and there is some new test code too. I can explain details on request, please try to have a look at the code, I have still many ideas for improvements :)

h3xstream:
I really liked this pull request. It is a long term implementation of taint detection. My previous implementation look very poor compare to this new system.
I also discover many FindBugs API along the code review.

I will remove the artifacts of the wobbly "Replay" mechanism I previously introduce. I will do minor contribution to thing like the list of transfer methods. Other than that, I don't see anything major change to come on my side. I will start to do some testing on large code base to see how it react.

Again.. You did a really great job!

h3xstream:
Mandatory reference to this related issue #14

formanek:
Changes are described in more detail here:
http://www.ysofters.com/2015/08/06/detection-of-hard-coded-passwords-improved/
http://www.ysofters.com/2015/08/31/taint-analysis-added-to-findbugs/

PR #66, opened 2015-07-26
Taint sources and improved taint transfer

formanek:


h3xstream:
Perfect. The split methods will be useful for eliminating some false positive regarding constant command being execute.

PR #67, opened 2015-07-27
Separated hard coded password and key reporting

formanek:
I have created a separated bug pattern for hard coded cryptographic keys - ConstantPasswordDetector now reports hard coded passwords for String and char array types and hard coded keys for the other types. I think this is appropriate since both weaknesses have different CWE ID and are equally important.


h3xstream:
Makes sens.

PR #69, opened 2015-08-01
Taint sources locations added to bug reports

formanek:
Taint analysis remembers all possible code locations with taint sources from now and shows them in bug reports (e.g. lines are highlighted in FindBugs GUI). If taint source is unknown and sink known, locations of unknown sources are shown. This makes reported warnings more informative and false positives can be distinguished more easily.


h3xstream:
Do IDE show those highlights (Eclipse/IntelliJ)?

formanek:
Only one line (with taint sink) is normally highlighted, but if the IDE plugin shows bug details (such as in IntelliJ), you can read the locations of taint sources (e.g. "At MyClass.java:[line 10]").

PR #70, opened 2015-08-02
Translate a message of HARD_CODE_KEY pattern.

naokikimura:


h3xstream:
Thanks for the contrib!

PR #75, opened 2015-08-25
Global taint analysis, improvements and bug fixes

formanek:
After many hours of implementing functionality and fixing problems, here is my new major contribution to taint analysis. The main improvement is an attempt to make taint analysis global, which is done by **automatic method taint summary generation** and by recursive **taint sink analysis**.

While a method is analyzed, taint states at the top of stack when visiting return instructions are merged and saved as a method summary. Taint objects now contain information about parameters contributing to the taint state, so it is possible to **infer transfer methods** too. Moreover, I have introduced something called non-parametric taint, which contains information about taint added inside a method, so the result can be something like *the method returns taint determined by the merge of parameters 1 and 2*, but also *the method taint state is unknown, but if the parameter 2 is tainted, then the result of the method is tainted too*. A derived summary can be used immediately in the method analysed next and since the methods are now analysed in call order, the depth of global analysis is unlimited (up to recursion and dependency cycles) while keeping good performance (every method is analysed once). These summaries can be checked by printing summary map (the format is the same as for configuration), I have been verifying it on a larger project and it seems to be working well (after fixing many problems), even though there are limitations (unknown mutable objects, inheritance polymorphism etc.) and more testing should be done. For example, it could properly analyse *query builders*, which use input parameters and constants to build query with taint state dependent on method inputs. One disadvantage is that all methods are now analysed (but only once as I said), so candidate method of injection sources is quite useless, but I think the performance is still good and these improvements worth some CPU time.

The second improvement making the analysis global is analysis of **taint sinks in separate methods**, which are called with tainted parameters. Here, we cannot use method summaries (they would have to be made for each taint sink, but sinks are different for each weakness type and not known during taint analysis). Instead, reporting of the detected bugs with medium confidence is **delayed** - if we found a tainted call, priority is raised, if not, it is reported at the end of the analysis. Methods containing sinks are stored and methods containing those methods too, so taint input can be found through more levels. There is an asymmetry between raising and lowering priority - if safe source is found for taint sink, we cannot report it just with low confidence, because there can be unknown tainted inputs, so this kind of analysis does not reduce reported false positives, but make more high confidence bugs reported.

I have improved displaying of locations, so you should see a **path between taint source and taint sink** even through several different classes. Sometimes, more locations were shown for one line, so I have made at most one location per line displayed.

I have also **improved local analysis**. Mixing strings with numbers in any form should not change the taint state to unknown from now. This is important, since numbers are often added into queries, so these false positives can be eliminated. I have added support for arrays of objects and some collections, so we can track the taint state in these containers. While implementing this, I found a really weird bug (I will made an issue for it soon), so I have restricted the analysis a bit and I hope it is stable now.

Last, but not least, there were a few **bugs** (that I have fixed), which could also affect previous versions and hide some true positives.
 
I missed many details in this description so feel free to ask about them. I appreciate any help with testing to be sure everything works correctly.


h3xstream:
I will look at this contribution tomorrow.

h3xstream:
The changes were all integrated.
I have some very minor suggestions regarding naming and structure now that I have a better understanding of the engine.

It would be great to document on the Wiki some concepts at high level. It is not urgent as it is rapidly evolving.

formanek:
I am writing blog articles about the improvements to http://www.ysofters.com/author/formanek/ (this pull request is not covered yet). I am going to write some javadoc later too.

h3xstream:
@formanek Perfect I'm going to link your articles in the wiki. It's an excellent doc for future developers (including general FindBugs dev).

Issue #76, opened 2015-08-31
EntityManager createQuery trips SECSQLIJPA even with safe usage


ThrawnCA:
Using the ``createQuery`` and ``createNativeQuery`` methods of a JPA EntityManager trips the SECSQLIJPA detector even if there are no tainted inputs.

This class trips the detector for me:

    package findbugs.sample;

    import javax.persistence.EntityManager;
    import javax.persistence.PersistenceContext;
    import javax.persistence.Query;

    public class TestInjectionWithConstantString {

        @PersistenceContext
        private EntityManager entityManager;

        public void queryWithConstantString() {
            entityManager.createQuery("foo");
            entityManager.createNativeQuery("foo");
        }

    }

h3xstream:
Yup. Confirmed. Note that the false positive are probably mark as LOW. (Contrary to Medium or High for Higher level confidence)
It was intended in the change done by @formanek.

I think I will be better to remove those false positives from the report.

h3xstream:
I'm planning to an urgent release of 1.4.3 ASAP.

formanek:
Yes, it is intended as an insurance in a case of a bad analysis, low confidence warnings are not shown in FindBugs GUI and I was inspired by h3xstream's code, when implemented this, so I have thought, it is ok. But if confidence levels are not supported by plugins, we should change it of course.

The easiest way of solving this is change of priority from low to ignore. Since I would like to have a simple way of checking, what is ignored, I think we could make a new debug rule (ignored by ordinary users) such as *Taint sink with safe sources* and report the bug instance here instead.

We should also consider, what to do with high priority bugs - I made an effort to prioritize high-probable bugs, but this information is lost, when priorities are not supported. This could be solved by duplicating bug patterns - e.g. *SQL Injection* for high confidence and *Potential SQL Injection* for medium confidence. The advantage of this would also be that the first rule should have a minimum of false positives and could be used in automatic analysis with alerts. (And the second during manual code reviews.) Another solution is just to add a tag *[high confidence]* to the report (Is the description visible in your tools?). What do you think? (@h3xstream, @ThrawnCA)

h3xstream:
@formanek : The fact that constant string are alerted is a bug/problem. Even in "Audit" mode, I don't want to see those.

The priority filter is applied to all bug patterns. I don't think it's possible to apply to a single bug pattern.

[...]

PR #78, opened 2015-09-02
Unknown methods made to modify taint state of their parameters to unknown

formanek:
Besides the String type, I have also excluded parameters of type java.lang.Object - these are not immutable, but cannot be modified without casting to a concrete type, so I think it is quite good approximation and better solution, then adding many methods to method-summaries.txt.

### DIFF
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/TaintFrameModelingVisitor.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/TaintFrameModelingVisitor.java
@@ -17,13 +17,13 @@
  */
 package com.h3xstream.findsecbugs.taintanalysis;
 
-import com.h3xstream.findsecbugs.common.ByteCode;
 import edu.umd.cs.findbugs.ba.AbstractFrameModelingVisitor;
 import edu.umd.cs.findbugs.ba.DataflowAnalysisException;
 import edu.umd.cs.findbugs.ba.InvalidBytecodeException;
 import edu.umd.cs.findbugs.ba.generic.GenericSignatureParser;
 import edu.umd.cs.findbugs.classfile.MethodDescriptor;
 import edu.umd.cs.findbugs.util.ClassName;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -57,6 +57,7 @@ public class TaintFrameModelingVisitor extends AbstractFrameModelingVisitor<Tain
 
     private static final String TO_STRING_METHOD = "toString()Ljava/lang/String;";
     private static final Set<String> SAFE_OBJECT_TYPES;
+    private static final Set<String> OBJECT_TYPES_NOT_TO_MODIFY;
     private final MethodDescriptor methodDescriptor;
     private final int parameterStackSize;
     private final TaintMethodSummaryMap methodSummaries;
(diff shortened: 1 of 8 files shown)


h3xstream:
This pull request will resolve #76 (Ref)

formanek:
@h3xstream  Well, I'm not sure about the Object class too, but do you know an example of a method, that modifies its Object parameter? If there is a method of some of the Object children, which modifies the instance passed as parameter, there should be an interface or superclass as a parameter type instead of Object.

Yes, primitive types are ignored. But arrays are ignored too, which is a mistake - I'll fix it in a minute.

The field was originally called IMMUTABLE_OBJECT_TYPES, but since I've added Object, I've renamed it to avoid confusion.

Methods can be removed, if we are going to keep Object excluded and if the mentioned methods do not modify any of their array parameters.

h3xstream:
@formanek For the Object being down cast and modify. I have two examples.

**Manual down casting**
```java
public void modify(Object data) {
   if(data instanceof Person) {
      Person p = (Person) data;
      p.setName("aaaa");
   }
}
```

**Implicit casting with the use of Generic**
```java
class DataProcessor extends GenericProcessor<Person> {
    public void modify(Person p) {
        p.setName("aaaa");
    }
}
```
The actual code/bytecode will be similar to:
```java
class DataProcessor extends GenericProcessor<Person> {
    public void modify(Object p) {
        ((Person) p).setName("aaaa");
    }
}
```

formanek:
@h3xstream The first example is not realistic for me - why wouldn't use more concrete type? I've tried the second example and even though a bridge method with Object parameter is generated, method *modify* has Person type when called on *PersonProcessor* from your example. Can you provide a code for *GenericProcessor* and a call of *modify* that has Object parameter?

h3xstream:
@formanek https://gist.github.com/h3xstream/0b02944e0017dcb50f54#file-output-result-L21

Some decompiler such as JD-gui will hide the Object type when they understand that Generics are being used.

formanek:
@h3xstream Ok, thank you. I was calling the method directly on the subclass, so it had the subclass type parameter.

However, I'm still not very convinced, that Object type should not be excluded. I think, If a parameter in a method of a superclass (or interface) is just Object and an instance of the extending (implementing) class is declared as its supertype (interface), then the parameter is not going to be modified. And even if it is modified, then not with a tainted value. I just cannot imagine a real code, that would do this. Do you know some examples from real API or can you just describe me the situation, where an Object parameter could be modified with a tainted value?

In contrast, when we include the Object type, logging API could often make a safe value unknown, since queries can be logged before executing and Object parameters are used here.

h3xstream:
Ok I see the point for the logging API. These are pretty common for logging query and such.
I would prefer **a white list of those API** to be more systematic. We could add them progressively.

I understand your point for quickly covering generic API such as logging, metric, etc. But, I prefer the certainty of the analysis over avoiding to maintain certain known safe methods.
`Object` can literally be anything. It is hard predict the impact on unknown API that people will develop.

One thing to keep in mind. The bytecode can also come from Groovy, Scala and even JSP. This support come for free because it is all bytecode. On the other hand, it is hard to predict that their respective compilers will not produce type erasure.

formanek:
Fine, Object type is now included. Methods *equals* are excluded from the mechanism and I have added Java logging API (not tested).

I had to uncomment two methods in method-summaries.txt. I have deleted those having just String arguments (and unimportant objects). The remaining disabled methods (starting with -) could be probably uncommented to improve analysis for collections and arrays - can you check, that there is no method, which taints some of its parameters?

Issue #80, opened 2015-09-07
Issue with constructor with List and array as parameter (Command injection detection)


h3xstream:
The following sample trigger vulnerability while it should be consider safe (false positive).
The issue is related to the constructor signature accepting `java.util.List` or `String` array.

```java
public class CommandInjectionSafe {

    public void safeEmptyConstructor() {
        new ProcessBuilder();
    }

    public void safeConstructorArray() {
        new ProcessBuilder("ls","-la");
        String[] cmd = new String[] {"ls","-la"};
        new ProcessBuilder(cmd);
    }

}
```

formanek:
@h3xstream This seems to be fixed after modifications, I have made.

Issue #90, opened 2015-09-14
Parametric taint state not changed when used as an argument of an unknown method


formanek:
When a parameter is passed to an unknown method, the taint state is changed to *unknown* if it was *safe* before. However, if there are taint parameters set, the non-parametric taint state was not modified. As a result, invalid method summaries could be generated. For example, the summary of a piece of code like this would be *0,SAFE* instead of *0,UNKNOWN*
```java
public List<String> test(List<String> list) {
    list.add("safe");
    modify(list);
    return list;
}
```

Issue #85, opened 2015-09-11
Bad method summaries derived for complex flow


formanek:
Because of the direct `Taint` instances modifications, the `Taint` class had its `equals` method defined only for the `state` attribute - full equals caused an error from the issue #82. For this reason, the analysis sometimes finished too early (with less iterations than actually needed) and some parameter was missing in the summary or non-parametric taint was invalid. Although this affected less than 1% of method summaries, it could cause both false positives and false negatives. This is fixed by not modifying values in frames directly and implementing proper `equals` method for `Taint`.

Issue #84, opened 2015-09-07
Invalid taint modifications of local variables, when loaded from method summary


formanek:
Local variable index of a taint added to a new method summary must be invalidated. Otherwise, invalid taint is loaded during method invocation and some unrelated local variables could be modified or even an invalid frame index accessed (throwing an exception).

PR #91, opened 2015-09-14
Improved taint analysis (several bugs fixed, refactoring)

formanek:
Analysis no longer relies on direct modifications of `Taint` instances in `TaintFrame` and the code is refactored to be more robust and clear. This pull request should fix issues #80, #82, #83, #84, #85 and #90. 

There is a change in constructor summaries - now there are two mutable stack indices, since new instances are duplicated before calling a constructor.


formanek:
Merging should be easy since I have merged your changes

h3xstream:
All changes were integrated.

I have added a test case for a miss LDAP injection a830a3d3ba30879b8de8c4a4dda85f519124cfef

PR #93, opened 2015-09-15
Added command injection sinks and CWE identifiers

### LINKED ISSUE #92: All Runtime.exec methods should be taint sinks
Only one-parameter `exec` methods are considered to be taint sinks for command injection. Tainted parameters in other methods are not detected, so there are false negatives.

formanek:
Fixes #92. CWE identifiers should help tools parsing XML output with results of analysis.


h3xstream:
Awesome

PR #94, opened 2015-09-15
Fixed incomplete candidate method for LDAP injections

formanek:
Added class constants, the test is switched on.


h3xstream:
AAAAAAAAAAAh! That's it! <-- My reaction the second opening the commit detail.

I spend so much time debugging the TaintDetector and TaintFrameModelingVisitor.

PR #107, opened 2015-10-09
Inheritance aware taint analysis, extended collections support

### LINKED ISSUE #98: Consider inheritance for method summaries
When a taint method summary is stored, the output taint should be transferred to its super methods by merging too. Vice versa, when a method summary is loaded and not found, we should search for the nearest super method. This would allow more accurate analysis and more concise method configuration. `Hierarchy2` class from FindBugs could be used in implementation.

### LINKED ISSUE #99: Better support for collections
Besides some types of lists, we should also support sets and maps and more implementing classes. Configuration of method summaries should not become too large after implementing #98. For full support, we also need to include iterators and methods from `Collections` class.

### LINKED ISSUE #106: Method arguments are not tainted if their derived summary is stored
Taint method summary derivation cannot derive taint changes in mutable stack indices and the summary is stored without change of arguments taint. Then the arguments are not tainted if this method is called, because taint summary exists. This could rarely cause false negatives. We have to differ between configured and derived summaries until we are able to derive mutable stack indices.

formanek:
Another evolution of the taint analysis closes #98, closes #99 and fixes #106.


h3xstream:
I didn't had the chance to review fully the commits. But it seems good as the refactoring support all the previous tests  and add new one.

High level review:
 - Not sure what this [assert](https://github.com/formanek/find-sec-bugs/commit/8cf145d783f65b5baf6391df642175606418a67b#diff-ad186a498d2c9c9e9a60e9c6df7266a8R57) is doing. Probably to avoid empty catch ?
 - With the high number of configuration, it would be nice to add tests to check the sanity of the configuration (method summaries).

formanek:
I am using ``assert false`` to indicate that something never happens (if I'm not mistaken). In this case, every implementation of Java platform is required to support UTF-8, so ``UnsupportedEncodingException`` should never be thrown. I am thinking of ``assert`` commands as inner documentation better than comments, because we are warned if the assertion doesn't hold (when running with ``-ea`` parameter), so I'm using these quite often.

Well, adding tests would be always nice, but just extending the configuration was quite time consuming and testing everything would be even worse. It's probably faster just to review them. I will rather implement new detectors and features...

h3xstream:
Perfect.
Just throwing an idea for the sanity test.

Issue #108, opened 2015-10-11
Add taint sinks for XPath injection


formanek:
Not all methods from ``org.apache.xpath.XPathAPI`` are included and there are more APIs to add.

formanek:
There is also a bug in ``XPathInjectionJavaxDetector`` - bad argument in method ``evaluate`` is checked (0 instead of 1). Plus the other 3 methods with different signatures are not checked at all.

h3xstream:
Consider fixed : https://github.com/formanek/find-sec-bugs/commit/5eb668675314bd2bd3e48f71167bb08a0fd4a941#diff-e7af01b3ebefa5f1b6ed5b5c44483905R1

PR #115, opened 2015-10-18
Refactored and fixed injection detectors

### LINKED ISSUE #96: Refactor injection detectors
I want to refactor detectors using taint analysis to use hash maps and full method names instead of confused and slow conditional statements. It should be very easy to add a new sink and check the existing sinks. I would also put another abstract class between `TaintDetector` and concrete detectors and make more methods abstract to enable alter the behaviour for concrete injection types for the future improvements. `InjectionSource` class can be removed - instances of `InjectionPoint` will be in the map, `isCandidate` method is obsolete. *Injectable method* need not to be set and can be replaced by calling methods from FindBugs `SignatureConverter`. We will also need to modify `CustomInjectionDetector`.

@h3xstream Do you agree? Can I start to implement this?

### LINKED ISSUE #97: Path traversal and Xpath injection detectors should use taint analysis
Current implementations use `OpcodeStackDetector`, which has very weak analysis for avoiding false positives. Creating taint detectors for these kinds of injection should be easy, especially after implementing #96.

### LINKED ISSUE #105: Missing taint sinks for LDAP Injection
Add coverage for the method search for DirContext
 - [DirContext.lookup()](http://docs.oracle.com/javase/7/docs/api/javax/naming/Context.html#lookup(javax.naming.Name))

Thanks to @mcwww for the information

### LINKED ISSUE #109: Missing taint sinks for Spring SQL injection
Besides missing package ``org.springframework.jdbc.core.simple.SimpleJdbcTemplate`` mentioned in source code, there are missing methods from ``JdbcOperations`` (and ``JdbcTemplate``) - namely 3 ``query`` methods and ``queryForList`` with ``Object...`` as last argument type and all ``update`` methods (accepting ``String``). Plus method [``newPreparedStatementCreator``](http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/jdbc/core/PreparedStatementCreatorFactory.html#newPreparedStatementCreator-java.lang.String-java.lang.Object:A-) from ``PreparedStatementCreatorFactory`` looks dangerously too.

### LINKED ISSUE #113: Path traversal false positives
There are two bugs in ``PathTraversalDetector``:

1. ``new RandomAccessFile("safe", "r")`` is reported, because access mode is checked (why?) for variable string and the condition is negated (typing error?), so a bug is shown for constants in the second argument. I think access mode should not be checked, since ``new RandomAccessFile("safe", tainted)`` is not very dangerous

2. ``new FileWriter("safe")`` is reported, because the checked argument is the same as for the 2-argument constructor, so the instance is checked (which is definitely not a constant string) and every ``FileWriter`` 1-argument constructor call is reported.

formanek:
Injection detectors are refactored and all use taint analysis (fixes #96, fixes #97), but an old approach is still supported with ``LegacyInjectionDetector`` and used for open redirect and custom injections. There were a few bugs and missing values in taint sinks (fixes #105, partly #108, fixes #109, currently disabled sinks for #111, fixes #113 and some more).


h3xstream:
I really like the `ConfiguredBasicInjectionDetector` and `BasicInjectionDetector` approach.

PR #122, opened 2015-10-23
Response splitting, hash functions and messages

### LINKED ISSUE #120: Add methods for weak message digest
The issue #119 made me to look at this detector and I have found 5 missing methods from [`DigestUtils`](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/digest/DigestUtils.html). We should also check using of weak hash functions in signature algorithms (class [`Signature`](http://docs.oracle.com/javase/8/docs/api/java/security/Signature.html)). I will add those methods and increase the reported priority.

### LINKED ISSUE #121: Detector for HTTP response splitting
I have created a taint detector for [HTTP response splitting](https://cwe.mitre.org/data/definitions/113.html) weakness. Potentially dangerous methods that I've included are `setHeader` and `addHeader` from `HttpServletResponse` and `setValue` and constructor from `Cookie`.

formanek:
Fixes #121 and fixes #120. In addition, I have modified many reported messages - injectable arguments are shown for every injection and messages should be more unified (capital letters, periods etc.), please check if there is something you don't like.

### DIFF
--- a/plugin-deps/src/main/java/javax/servlet/http/HttpServletResponse.java
+++ b/plugin-deps/src/main/java/javax/servlet/http/HttpServletResponse.java
@@ -8,6 +8,8 @@ public interface HttpServletResponse extends ServletResponse {
     void addCookie(Cookie cookie);
 
     void addHeader(String header, String value);
+    
+    void setHeader(String header, String value);
 
     void sendRedirect(String url) throws IOException;
 }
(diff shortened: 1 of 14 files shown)


h3xstream:
The HTTP response splitting rule will probably be noisy for must developer. The proper fix not related to the implementation but more about having the latest web container (tomcat, jetty, glassfish, etc.)
It can still be useful for a security auditor that could test the web container for the specific cases found.

I would make the rule trigger with Medium if confirm to use tainted input and with Low if the exploitability is not confirm.

PR #136, opened 2015-12-04
Properly handle paths to files

jsotuyod:
- Spaces and other chars would get encoded hen creating the URL,
    this fix allows to get a decoded path that can actually map to the filesystem.

### DIFF
--- a/findbugs-test-util/src/test/java/com/h3xstream/findbugs/test/service/ClassFileLocator.java
+++ b/findbugs-test-util/src/test/java/com/h3xstream/findbugs/test/service/ClassFileLocator.java
@@ -17,9 +17,11 @@
  */
 package com.h3xstream.findbugs.test.service;
 
+import java.net.URISyntaxException;
 import java.net.URL;
 
 import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.fail;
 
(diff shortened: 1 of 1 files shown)


h3xstream:
Thanks I'll will trust you on this one. I didn't know about the diff between `toString()` and `toPath()` methods.

h3xstream:
Quick question: Are you using the test utility for your own detectors ?
I will start to consider to keep some backward compatibility if so..

jsotuyod:
@h3xstream yes we do! https://github.com/Monits/findbugs-plugin uses this for our own unit testing (it's awesome!), that's why we have submitted several PRs over time.

In this case, we realized that if our working directory was something like "/home/jsotuyod/workspace/findbugs plugin/", we would get paths such as "/home/jsotuyod/workspace/findbugs**%20**plugin/target/test-classes/samples/system/UseSystemErr.class" which of course would not be resolved.

The [official doc on URL](http://docs.oracle.com/javase/7/docs/api/java/net/URL.html) says

> Note, the URI class does perform escaping of its component fields in certain circumstances. The recommended way to manage the encoding and decoding of URLs is to use URI, and to convert between these two classes using toURI() and URI.toURL().

PR #157, opened 2016-01-07
Fix typo in taint-config/java-lang.txt

apasel422:


h3xstream:
Automate testing for typo #158

PR #166, opened 2016-01-14
Taint tags - support for taint sanitization (strating with XSS)

formanek:
First version of sanitization support - methods can be configured to add or remove tags for method output and detectors can read them to decide whether to report the bug. Currently, only XSS_SAFE tag is supported, but I'm going to add more before the new release.

Sanitized values for XSS are reported with low priority (less than for inputs with unknown taint state). I don't want to exclude those detections completely, because configured escape methods are context specific and there is still a risk of a successful XSS if an inappropriate sanitization method is chosen.

### DIFF
--- a/plugin-deps/src/main/java/org/owasp/esapi/Encoder.java
+++ b/plugin-deps/src/main/java/org/owasp/esapi/Encoder.java
@@ -2,4 +2,5 @@ package org.owasp.esapi;
 
 public interface Encoder {
     String encodeForHTML(String input);
+    String decodeForHTML(String input);
 }
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/injection/BasicInjectionDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/injection/BasicInjectionDetector.java
@@ -21,7 +21,6 @@ import edu.umd.cs.findbugs.BugReporter;
 import edu.umd.cs.findbugs.util.ClassName;
 import java.util.HashMap;
 import java.util.Map;
-
 import org.apache.bcel.generic.ConstantPoolGen;
 import org.apache.bcel.generic.InstructionHandle;
 import org.apache.bcel.generic.InvokeInstruction;
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/injection/SinksLoader.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/injection/SinksLoader.java
@@ -1,11 +1,27 @@
+/**
+ * Find Security Bugs
+ * Copyright (c) Philippe Arteau, All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3.0 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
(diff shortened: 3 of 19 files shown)


h3xstream:
Again excellent contrib. !
I will take some time to see if the score increase for OWASP Benchmark.

aabashkin:
Nice job, thank you for the work

Anybody interested in tainting should review the Checker framework implementation : http://types.cs.washington.edu/checker-framework/current/checker-framework-manual.html#tainting-checker

"@Tainted and @Untainted can be used for many purposes

The @Tainted and @Untainted annotations have only minimal built-in semantics. In fact, the Tainting Checker provides only a small amount of functionality beyond the Subtyping Checker (Chapter 22). This lack of hard-coded behavior means that the annotations can serve many different purposes. Here are just a few examples:

* Prevent SQL injection attacks: @Tainted is external input, @Untainted has been checked for SQL syntax.
* Prevent cross-site scripting attacks: @Tainted is external input, @Untainted has been checked for JavaScript syntax.
* **Prevent information leakage**: @Tainted is secret data, @Untainted may be displayed to a user.
In each case, you need to annotate the appropriate untainting/sanitization routines. This is similar to the @Encrypted annotation (Section 22.2), where the cryptographic functions are beyond the reasoning abilities of the type system. In each case, the type system verifies most of your code, and the @SuppressWarnings annotations indicate the few places where human attention is needed."


I am **very** interested in the last use case (information leakage). Other problems such as XSS & SQL already have some good solutions available. What if a use case just involves a @SensitiveString or a @SensitiveDomainObject that the application is trying to protect? It should never make it to console.log, JspWriter, HttpClient, or other common sinks.

Can the current XSS taint solution be generalized to include something of this sort?

h3xstream:
@aabashkin 

     Can the current XSS taint solution be generalized to include something of this sort?

Short answer, yes it already is. Taint analysis is not new to FSB.

A proper taint analysis is implemented since 1.4.2. It was contributed by @formanek. It continue to improved on each release. It is used by all injections detectors, XSS detectors, open redirect, etc.

Information leakage detection will most likely use the tagging mechanism.

formanek:
@h3xstream Thanks. I'm afraid this will not improve the score in OWASP Benchmark, because the proper detection will still be blocked by unrealistic constructions like expressions that are always true, but more taint tags should improve score in Juliet Test Suite (which I consider to be more useful benchmark).

@aabashkin I'm thinking of using taint analysis also for hard coded passwords / keys or information leakage in the future, it should be quite easy with tags, but injections are my priority for now. One problem is that sources of sensitive information are not so easy to identify automatically and I think not many users are willing to add annotations into their code to improve the detection.

aabashkin:
I took a look at https://github.com/find-sec-bugs/find-sec-bugs/releases/tag/version-1.4.2 and now I understand more clearly, thanks

http://www.ysofters.com/2015/08/31/taint-analysis-added-to-findbugs/ was very useful as well

h3xstream:
@formanek I agree for OWASP Benchmark. I also think that the important stuff is already cover. I had in mind few TN test cases including:  
https://github.com/OWASP/Benchmark/blob/b939c34e6cd79c089a6ed38d806146c43b521dd8/src/main/java/org/owasp/benchmark/testcode/BenchmarkTest00098.java#L56
The test case was missed because the value push to the session attribute is escaped for XSS. It was previously marked as SAFE.
Other than that, I don't think their is more attention needed for the project.

PR #171, opened 2016-02-03
Support for sanitization using replace methods in String

formanek:
`Taint` instances are now remembering the constant value (even for char) and you could use it instead of `Constant` in `ConstantDataflow` for Strings. This is used to check parameters of replace methods in `String` and add taint tags to the return value, so we are able to check it in taint detectors. Since this is not the best way to sanitize inputs, priority is set to low (not ignore) for SQL injection and XSS, when tags from replace methods are used. CR and LF replacement should be enough for CRLF injection, so priority is set to ignore in that case.


h3xstream:
https://github.com/formanek/find-sec-bugs/commit/c7f474d2b82b4e704eaa4a7111cdde42cf7f6c2d#diff-c1d98326cbc7a19685ad97d009b12372R405

The support for `.replace("\n","")` is pretty nice.
I think it would be nice to generalize the concept of dynamic method summaries.

Another exemple is dynamic JSTL expression in JSPs.
`createValueExpression(_jspx_page_context.getELContext(), "${param.expression}", Object.class)`
The expression could be safe if a sanitizing function is call inside the expression..

PR #173, opened 2016-02-14
Better sink confirmation mechanism, safe fields

formanek:
The mechanism, which is responsible for increasing priority when a method containing an injection sink is called with tainted argument (linking to the sink inside), was not very reliable - tests passed, but for nested sinks it sometimes worked only by chance and I wasn't able to extend functionality (old tests failed strangely). For this reason, I rewritten the mechanism.

The priority can be now decreased too if there are only safe calls. Because it's not easy to detect all kinds of method usage (polymorphism, reflection etc.) and we might want to detect bugs in incomplete software too, the priority (or confidence) is still medium when reporting unused methods. But if the method is called with some safe argument (linked to sink) and there are only safe calls (including sanitized inputs), the priority is changed to low. This could remove false positives if tainted inputs are not sanitized directly before usage but earlier or the method with sink is not intended to be called with user input. Comment with information about which usage was detected is included into the bug instance.

The whole problem is quite complex considering all the edge cases, so I'm still not sure that everything works correctly now. One issue I know about is that sometimes a sink called using polymorphism is not confirmed even though the analysis should support it. This is not deterministic - if I remove some unrelated files from analysis, it works. The same issue occurred also before changing the mechanism and I wasn't able to find the cause, it might be in the FindBugs API too (e.g. bad ordering of methods for many classes to analyse). I'm not going to spend much time solving this, because it happens only sometimes and the bug is still reported with medium priority.

Another improvement I added is that reading a class attribute with safe type now stores a safe value of taint instance. More tainted locations should be added to the report and duplicate lines should be eliminated.


formanek:
No test failed and I see no exceptions in Travis CI, but
```
|INFO | FindBugsLauncher : Analyzing... /home/travis/build/find-sec-bugs/find-sec-bugs/plugin/target/test-classes/bytecode_samples/scala_command_injection.jar
Killed
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.18.1:test (default-test) on project findsecbugs-plugin: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.18.1:test failed: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
[ERROR] Command was /bin/sh -c cd /home/travis/build/find-sec-bugs/find-sec-bugs/plugin && /usr/lib/jvm/java-8-oracle/jre/bin/java -jar /home/travis/build/find-sec-bugs/find-sec-bugs/plugin/target/surefire/surefirebooter531172736335644084.jar /home/travis/build/find-sec-bugs/find-sec-bugs/plugin/target/surefire/surefire7295245775214118369tmp /home/travis/build/find-sec-bugs/find-sec-bugs/plugin/target/surefire/surefire_02951700618737220530tmp
```
What could be the problem? Local build finishes successfully.

formanek:
Fixes #176, fixes #177 and Travis seems to be ok now

h3xstream:
Sorry for the delay

h3xstream:
I opened an issue #180 regarding the build failing .. I don't know why it is failing on travis-ci.

PR #174, opened 2016-02-15
Fix URL shown for CUSTOM_INJECTION bug warning

mkienenb:
http://h3xstream.github.io/find-sec-bugs/bugs.htm#CUSTOM_INJECTION redirects to http://find-sec-bugs.github.io/bugs.htm and not to http://find-sec-bugs.github.io/bugs.htm#CUSTOM_INJECTION.   Since this is part of the find-sec-bugs site now, it makes more sense to directly use the find-sec-bugs.github.io URL.


h3xstream:
For some reason Travis-CI is failing.. I will take a chance on this one .. since it's just a URL change.

Be aware, that the warning will disappear by default in the next release.
Are you using custom APIs ? I am thinking about improving the integration for the custom rules configurations.

mkienenb:
No, I am not using custom APIs, at least not yet.   I think having the warning disappear by default is a good idea.

PR #178, opened 2016-02-16
Add Struts DynaValidatorForm support in addition to ValidatorForm

mkienenb:
False Positive for SECSFV: STRUTS_FORM_VALIDATION: Struts Form without input validation 

A DynaValidatorForm doesn't extend from ValidatorForm but still has a validate method.


h3xstream:
Thanks

PR #185, opened 2016-04-19
Correct japanese messages formatting

marcosbento:
Here are some minor fixes to the japanese messages. These corrections are mostly related to encoding of HTML tags in  <Details> section.


h3xstream:
Hum.. the build seems to be failing because of memory issue. (Unrelated to your change)

h3xstream:
Thanks again

PR #187, opened 2016-04-30
New analysis parameters and extended taint config

formanek:
Specifying `findsecbugs.taint.taintedsystemvariables=true` cause system variables to be considered tainted, `findsecbugs.taint.customconfigfile=path-to-my-config.txt` allows rewriting taint config in general and  `findsecbugs.taint.taintedmainargument=false` cause considering the argument of the main method to be safe (it is tainted by default).

I have added configuration for `Scanner` and `JTextComponent` (Swing GUI), I am still not sure about some servlet methods. I've also added `setHeader` as a taint sink for Unvalidated Redirect.


formanek:
@h3xstream Do you have an idea, why Travis is still failing?

h3xstream:
@formanek The process running the test is being killed.. Most likely an unresolved memory issue.

PR #188, opened 2016-05-11
Update owasp.txt

s-tikhomirov:
Fix error: encodeFromURL --> decodeFromURL
https://owasp-esapi-java.googlecode.com/svn/trunk_doc/2.0-rc4/org/owasp/esapi/Encoder.html#decodeFromURL(java.lang.String)


formanek:
Thanks

h3xstream:
I'm not sure about the purpose of this modification.

[encodeForURL](https://owasp-esapi-java.googlecode.com/svn/trunk_doc/2.0-rc4/org/owasp/esapi/Encoder.html#encodeForURL(java.lang.String)) is expected to be safe from XSS in the specific context of URLs.

Example:
```java
String html = "<a href=\"http://"+Encoder.encodeFromURL(req.getParameter("input"))+"\">Click me</a>"
```

The decode on the other hand does not provide any protection regarding escaping..
```java
String html = "<a href=\"http://"+Encoder.decodeFromURL(req.getParameter("input"))+"\">Click me</a>"
```

formanek:
@h3xstream the `-` sign before the tag means that tag is removed by the method, so the value is considered insecure after decoding even if it was encoded before.

h3xstream:
@formanek Got it ! Thanks

PR #190, opened 2016-05-24
Safe enums, dates, time and context path + javadoc

formanek:
Improves analysis a bit.


h3xstream:
Excellent. You took some advance on me.

PR #202, opened 2016-07-11
CWE 200 (Information Exposure) - Scala Play vulnerable code

MaxNad:
Detector for information exposure in Scala Play.

If a controller returns a variable containing the result of a tainted call to the configuration manager, an attacker could use it to obtain sensitive data like the application key or the database credentials.

Taints can now be annotated with the "SENSITIVE_DATA" tag to allow a better scalability of the detector.

I also added multiple taint definitions for data structures in Scala, this will facilitate the development of injection based detectors in Scala Play.


h3xstream:
Cool idea.

This include a huge block of method behavior (method summaries) for Scala string manipulation. This is a very good thing for all taint related analysis in Scala.

(I am reviewing the changes)

plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/TaintFrameModelingVisitor.java
```diff
@@ -26,13 +26,9 @@
...
 import edu.umd.cs.findbugs.util.ClassName;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Set;
+
+import java.util.*;
```
h3xstream: Please configure your IDE to avoid wildcards. (Minor issue)

For IntelliJ: http://stackoverflow.com/a/3348855/89769

plugin/src/main/resources/metadata/findbugs.xml
```diff
@@ -169,6 +170,7 @@
     <BugPattern type="OBJECT_DESERIALIZATION" abbrev="SECOBDES" category="SECURITY" cweid="502"/>
     <BugPattern type="DESERIALIZATION_GADGET" abbrev="SECDESGAD" category="SECURITY" cweid="502"  experimental="true"/>
     <BugPattern type="TRUST_BOUNDARY_VIOLATION" abbrev="SECTBV" category="SECURITY" cweid="501"/>
+    <BugPattern type="SCALA_SENSITIVE_DATA_EXPOSURE" abbrev="SECSDATALEAK" category="SECURITY" cweid="200"/>
```
h3xstream: The abbreviation could be shorter. This is not intended to be readable. Could be "SECSDL". (Minor issue)

It's a good thing to add the Scala prefix. This is an idea that should be applied to Java and other APIs. (Comment)

plugin/src/main/resources/metadata/messages.xml
```diff
@@ -3799,4 +3799,45 @@ Path traversal <sup>[3][4]</sup> are not possible.
     </BugPattern>
     <BugCode abbrev="SECXSLT">A malicious XSLT could be provided</BugCode>
 
+    <!-- XSLT -->
```
h3xstream: This comment that separate the description should be change to "Scala data leak" or something equivalent.

plugin/src/test/java/com/h3xstream/findsecbugs/scala/ScalaSensitiveDataExposureDetectorTest.java
```diff
@@ -0,0 +1,86 @@
...
+        methodFalsePositiveLines.put("safeUntainted", new int[]{105, 106, 107, 108, 109, /**/ 111, 112, 113, 114, 115});
+        methodFalsePositiveLines.put("safeNoSensitiveData", new int[]{119, 120, 121, 122, 123, /**/ 125, 126, 127, 128, 129});
+
+        for (Entry<String, int[]> entry : methodFalsePositiveLines.entrySet()) {
+            // Lets check every line specified above
+            for (int line : entry.getValue()) {
+                verify(reporter,never()).doReportBug(
+                        bugDefinition()
+                        .bugType("SCALA_SENSITIVE_DATA_EXPOSURE")
+                        .inClass("SensitiveDataExposureController").inMethod(entry.getKey()).atLine(line)
```
h3xstream: `atLine()` is optional. If it is omit, it is become "any line". The bug definition follow the Builder pattern. All attributes are optional.

`verify(reporter, never()).[...].atLine(1)` or 
`verify(reporter, never())`
will have the same behavior.

On the other hand, the test has higher chance of becoming false negative (placebo).
For example, if the detector gets broken and the sample is refactored (lines of code changed). It will silently pass.

h3xstream:
Great contribution !

@MaxNad I have left 4 comments. If you can push the 4 small changes, I will merge the change.

MaxNad:
I have pushed the modifications.

Thanks for the comments.

h3xstream:
Perfect! Thanks again.

I will wait for Travis-CI before merging.

PR #203, opened 2016-07-12
XSS (CWE 79) - Scala Play vulnerable code

MaxNad:
Detector for information exposure in Scala Play.

Using unescaped user input in web pages generation could lead to fishing, session hijacking or other harmful things.

This detector checks for basic XSS in Twirl and MVC API for Scala Play. It does not check for .scala.html files just yet.

Also added the AbstractInjectionDetector.getPriorityFromTaintFrame method to allow more modular injection based detectors. This will allow to check for a combination of parameters (for instance a tainted object calling a method with a specific value) - I needed this to check for MvcResult.as("text/html")

I have renamed the Sensitive data exposure injection sink file to match the naming convention of the other sinks.


MaxNad:
I do think this will cover item "XSS in Play controller : Ok().as("text/html")" in #111

MaxNad:
Since this pull request is still open, I took this opportunity to fix a message in the message.xml file. The vulnerable code snippet in SCALA_COMMAND_INJECTION was a safe one.

plugin/src/main/java/com/h3xstream/findsecbugs/scala/XssMvcApiDetector.java
```diff
@@ -0,0 +1,94 @@
...
+            Taint mvcResultTaint = fact.getStackValue(1);
+
+            // The MVC Result object was tainted - This could still be safe if the content-type is a safe one
+            if (!mvcResultTaint.isSafe()) {
+
+                // Get the value of the content-type parameter
+                Taint parameterTaint = fact.getStackValue(0);
+
+                if (!parameterTaint.isSafe()
+                        || parameterTaint.getConstantValue().equals(VULNERABLE_CONTENT_TYPE)) {
```
h3xstream: `parameterTaint.getConstantValue()` could be null ?

h3xstream: `text/html` is perfect. Here is an additional idea for more advanced vulnerabilities.

This same detector could be use to detect XSS through content-sniffing. `text/plain` and `application/octet-stream` are some of the Content-type to cover.

plugin/src/main/resources/metadata/messages.xml
```diff
@@ -3840,4 +3840,90 @@ Path traversal <sup>[3][4]</sup> are not possible.
...
+<p>
+    <b>Vulnerable Code:</b>
+<pre>def doGet(value:String) = Action {
+    Ok("Hello "+value+" !").as("text/html")
+  }</pre>
+</p>
+<p>
+    <b>Solution:</b>
+<pre>def doGet(value:String) = Action {
+    Ok("Hello "+value+" !").as("text/html")
```
h3xstream: The solution identical to the vulnerable code.
I would mentions **OWASP Encoder** (`Encode.forHtml()`) and **Common Lang** `StringEscapeUtils.escapeHtml4()` librairies. I don't know any Scala one.

- org.owasp.encoder.Encode.forHtml()
- org.apache.commons.lang3.StringEscapeUtils.escapeHtml4()

h3xstream:
I merged the change.
I have created ticket for remaining task #207 and new idea #205.

The only that I didn't create a ticket for is the change in the solution code sample.

plugin/src/main/resources/metadata/messages.xml
MaxNad: This one slipped past me.

Scala sends it as plain test if there is not .as() call.
Should I change the solution to : Ok("Hello "+value+" !") or use an encoder ?

h3xstream: `plaint/text` content-type could be interpreted as HTML in certain browser. So the later : encoder.

h3xstream:
@MaxNad I added your name to the contributors list : https://github.com/find-sec-bugs/find-sec-bugs#notable-contributions

Issue #182, opened 2016-03-16
INSECURE_COOKIE detector can be fooled by creating two or more cookies


jborgland:
The detector seems to be mixing up information from all cookies created in one method. Example below

This generates two instances of INSECURE_COOKIE (for both c1 and c2):
```java
  public void addCookies(HttpServletResponse response) {
    Cookie c1 = new Cookie("c1", "foo");
    c1.setSecure(false);
    response.addCookie(c1);
    
    Cookie c2 = new Cookie("c2", "bar");
    c2.setSecure(false);
    response.addCookie(c2);
  }
```

While this generates no instance at all (_although one would be expected, for c1_):
```java
  public void addCookies(HttpServletResponse response) {
    Cookie c1 = new Cookie("c1", "foo");
    c1.setSecure(false);
    response.addCookie(c1);
    
    Cookie c2 = new Cookie("c2", "bar");
    c2.setSecure(true);
    response.addCookie(c2);
  }
```

h3xstream:
Thanks for the report. It is was a known limitation. Minimal effort was put to avoid FP/TN for this rule.

But it the missed cases are not esoteric at all, so I'll put some effort to fix this. (I don't know when)

MaxNad:
I think I found a way to correctly handle this case while trying to create the Scala version of this detector.

You can see the fix here : https://github.com/MaxNad/find-sec-bugs/tree/issue_182
I will cleanup the code a bit and see if this is good enough to create a pull request.

PR #204, opened 2016-07-14
Potential fix for issue #182 (INSECURE_COOKIE detector can be fooled by creating two or more cookies)

MaxNad:
I changed the behavior of the CookieFlagsDetector and it now covers the example found in bug #182 .

It now behaves as a "mini taint detector", some of the default Find-Bugs checks are working like this (like the SqlInjection one) and they seem to work fine.

It first finds the cookie creation call, saves the position of the cookie on the stack and checks if it can find the .setHttpOnly or setSecure method for this object on the stack.

### DIFF
--- a/plugin/src/test/java/com/h3xstream/findsecbugs/cookie/CookieFlagsDetectorTest.java
+++ b/plugin/src/test/java/com/h3xstream/findsecbugs/cookie/CookieFlagsDetectorTest.java
@@ -52,7 +52,7 @@ public class CookieFlagsDetectorTest extends BaseDetectorTest {
         }
 
         // Advanced checks when multiple cookies are set
-        List<Integer> lines = Arrays.asList(new Integer[] { 72, 76, 80 });
+        List<Integer> lines = Arrays.asList(new Integer[] { 72, 76, 80, 84 });
         for (int line : lines) {
             verify(reporter).doReportBug(
                     bugDefinition()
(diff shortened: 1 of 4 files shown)


plugin/src/test/java/com/h3xstream/findsecbugs/cookie/CookieFlagsDetectorTest.java
```diff
@@ -84,18 +84,15 @@ public void avoidSecureFlagBasicFalsePositive() throws Exception {
...
-        );
-        verify(reporter,never()).doReportBug(
-                bugDefinition()
-                        .bugType("HTTPONLY_COOKIE")
-                        .inClass("HttpOnlyCookieSamples").inMethod("multipleCookies").atLine(79)
-                        .build()
-        );
+        List<Integer> lines = Arrays.asList(new Integer[] { 68, 79, 88 });
+        for (int line : lines) {
+            verify(reporter, never()).doReportBug(
```
h3xstream: Avoid using **never()** with the property **line** set. The broken tests could be badly fixed with new lines and stilll pass.

Most tests are using this pattern:
 - Validate true positive first. (with the lines specify)
 - Validate the count of true positive (`verify(reporter,times(X)).bugType("BUG_X").inClass("Class").inMethod("method").build()`)

MaxNad: Noted.

h3xstream:
I will accept the change.

On the long term, we will need to generalize the pattern trough a utility class or a base Detector.
There few detectors where we need to track instances : cookie attribute, XXE and key size for cipher.

PR #208, opened 2016-07-18
XSS (CWE 79) - Scala - The detector can be fooled when the .as("text/html") is in uppercase

MaxNad:
I fixed the "solution" part of the SCALA_XSS_MVC_API message in message.xml (#203).

I also realized that .as("TEXT/HTML") was not properly detected. I changed the detector to ignore case during his check.


plugin/src/main/java/com/h3xstream/findsecbugs/scala/XssMvcApiDetector.java
```diff
@@ -47,8 +47,8 @@ protected int getPriorityFromTaintFrame(TaintFrame fact, int offset)
             // Get the value of the content-type parameter
             Taint parameterTaint = fact.getStackValue(0);
 
-            if ( parameterTaint != null
-                    && (!parameterTaint.isSafe() || parameterTaint.getConstantValue().toLowerCase().equals(VULNERABLE_CONTENT_TYPE)) ) {
+            if ( !parameterTaint.isSafe()
```
h3xstream: of simply : 
`if(!parameterTaint.isSafe() || VULNERABLE_CONTENT_TYPE.equalsIgnoreCase(parameterTaint.getConstantValue())) {`

h3xstream:
I almost lost track of this Pull Request

PR #219, opened 2016-09-06
Add configurable custom xss detector

kuhnmi:
Recently, changes that allow a user to configure detectors on a per class basis have been merged into find-sec-bugs. I've tried to use that to register some custom XSS sinks. As opposed to custom-injection sinks (registered with the `CustomInjectionDetector`), they should only trigger if the input is _not_ html escaped. With the existing detectors that didn't seem possible, because there is no generic XSS detector.

This PR thus unifies code common to all XSS detectors in a new abstract class `BasicXssInjectionDetector` and adds a new generic `CustomXssInjectionDetector` subclass that can be used to register custom XSS sinks. It also adds a new bug definition `CUSTOM_XSS_INJECTION`, and a test that verifies a custom sink only triggers if injected parameters are not html escaped.

I hope I've roughly met the style of the project, and this stuff sounds useful to you. If not, please let me know.

### DIFF
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/xss/BasicXssInjectionDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/xss/BasicXssInjectionDetector.java
@@ -1,0 +1,52 @@
+/**
+ * Find Security Bugs
+ * Copyright (c) Philippe Arteau, All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3.0 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library.
+ */
+package com.h3xstream.findsecbugs.xss;
+
+import com.h3xstream.findsecbugs.FindSecBugsGlobalConfig;
+import com.h3xstream.findsecbugs.injection.BasicInjectionDetector;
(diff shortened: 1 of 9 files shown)


kuhnmi:
I've just noticed that the test that checks escaped parameters do not trigger a warning was not verified. I've added the corresponding verification.

h3xstream:
The main gain from using this implementation over custom injection configuration is having bug pattern ID/description explicitly telling it's an XSS.

I think the custom injection configuration files should be modified to support the configuration of arbitrary bug patterns. 🤔

kuhnmi:
Sorry for my long silence. The main benefit here is actually that a custom XSS detection does _not_ lead to a warning (i.e. has priority `LOW_PRIORITY`) if the corresponding injection is tagged as `XSS_SAFE` (or is appropriately encoded). That's in the part of the code that moved to `BasicXSSInjectionDetector`. I agree it would be nicer to solve this problem in a more generic way, i.e. to make `CustomInjectionDetector` able to deal with taint tags. (As far as I understand the existing code, it's currently not able to do so, please correct me, if I'm wrong.) We just needed this functionality in our project for XSS, thus I've created a fork and implemented it for the specific case of custom XSS injections. 

I agree, as a project owner, I'd probably also prefer a more generic solution, that makes `CustomInjectionDetector` able to properly deal with taint tags. I guess a first step would be to better reuse the code around taint tags (note the duplicated code that deals with XSS tags that moved to a single base class in my PR.) If not using inheritance (as in my PR), but composition, it should afterwards not be too hard to make `CustomInjectionDetector` (and other detectors) reuse the corresponding code. Unfortunately, at the moment I don't have time to work on a more complete solution than this quick hack for XSS.

I'd thus suggest to close this PR and instead open a feature request about making `CustomInjectionDetector` able to work with taint tags. But I leave this decision to you.

NB: Thanks for the great project!

h3xstream:
I have not fully investigate but it might already be possible to use this system variables: 
https://github.com/find-sec-bugs/find-sec-bugs/blob/1d288ef15122a4d883343769dd221cbe7bbeecb1/plugin/src/main/java/com/h3xstream/findsecbugs/injection/BasicInjectionDetector.java#L100

h3xstream:
Since this PR does not add new capabilities. I will not be integrated.

Issue #211, opened 2016-07-27
Implement a base detector to track specific calls


MaxNad:
Create a base detector to track specific calls on objects.

This enhancement proposition is based on the comment in #204 and should help fix #182.

The detector should be able to detect the following :

     Cookie item = new Cookie("f", "bar");	← Save the index of this object to track calls
     item.setSecure(false);					← Call to setSecure has the wrong value (should be true)
     										← Missing call to setHttpOnly(true)

MaxNad:
@h3xstream :

I started working on a base detector (https://github.com/MaxNad/find-sec-bugs/tree/instance_tracking_detector)

For now, I am using the cookie use case. If you have any more or if you have some ideas for this detector, feel free to send them to me.

Issue #201, opened 2016-06-29
New Rule: Use of setEscapeModelStrings in Wicket project


h3xstream:
Ref: http://stackoverflow.com/questions/1804220/how-to-display-html-text-within-a-wicket-element

PR #220, opened 2016-09-07
Implement a base detector to track specific calls (#211)

MaxNad:
This pull request is based on the enhancement proposition in #204.

The BasicInstanceTrackingDetector can be used to track specific Invoke calls on object instances.

When this detector is extended, you just have to specify the object instanciation instruction, the tracked calls and the reporting behavior. The detector allows you to track multiple instanciation instructions and calls (For instance Cookie.<init> and Library.CreateCookie) could be used to track an object creation).

I also changed the CookieFlagsDetector to test this new Abstract detector and added some more tests cases that the old version could not correctly identify.

This Abstract detector is based solely on the cookie flags use case. Do not hesitate to ask for changes or to suggest improvements.


h3xstream:
FYI: I might **not have the time** to take a look at integrating your PR next week.

MaxNad:
Take your time.
This PR doesn't change much to the actual behavior of the CookieFlagDetector.

h3xstream:
The main objection I have with this approach is that it support the Cookie special case. Yet it is not a "tool" that could be reuse easily.

XXE rules can't benefit from this implementation.
The configurations are not on the same instance has the instance triggering the vulnerabilities.

I have a feeling the taint analysis can be extends to support these scenario.

h3xstream:
This bug pattern #201 is pretty similar to the cookie bug patterns.

h3xstream:
@MaxNad I now have clearer idea of what could be done.

Reusing the taint analysis, we could allow detector to subscribe to specific events such as method calls.
When those events are trigger, detector can add their own analysis to the mix such as adding [Custom Tags](https://github.com/find-sec-bugs/find-sec-bugs/blob/master/plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/Taint.java#L89). Adding values to the enum of tag whould be the only modification to the core.

Example for Cookie analysis:

```java

//1. First, a constructor is called
cookie = new Cookie("test1","1234"); //--> The detector add a tag HTTP_COOKIE for the value
//2. Later..
cookie.setSecure(false); //--> The detector add a tag HTTP_COOKIE_SECURE 
//3. Final analysis
//In this case, it would be the end of the method or when a local value is overridden.
//The analysis would be look for local that have the tag HTTP_COOKIE **but, not** HTTP_COOKIE_SECURE or HTTP_COOKIE_HTTPONLY. 
```

@formanek If you see any problems: obstacle or obvious performances issue, let us know..

MaxNad:
This sounds great.
I will take a look at it as soon as I am done with my finals.

formanek:
Yes, I was designing taint analysis so that *taint tags* could be used for detection of problems other than injections in the future. Taint configuration is quite generic and allows to add taint tags for specific calls (with specific arguments). I'm just afraid that tracking of boolean values is not currently supported, but it shouldn't be difficult to implement. Adding tag HTTP_COOKIE shouldn't be needed, because tags are added only for specific classes plus there are other ways to find out the type of variables later.

Important thing to note is that detectors only __use the result__ of performed taint analysis without modification of `Taint` instances. So the correct way of implementing a new detector based on taint analysis is:

1. Improve taint analysis (by making it more general, not more specific) so it is powerful enough to handle the problem
2. Add custom taint tag(s) to the `Taint` class
3. Create a new file in taint-config describing the logic of taint transfer and load it in taint analysis
4. Extend a detector in taint-injection detectors hierarchy and implement reporting of the *sinks*

h3xstream:
@formanek Good point! The taint config could be used to add tag for some cases.

MaxNad:
@h3xstream, @formanek 
Just making sure I understand correctly.

The goal is to improve the taint analysis and create a new detector that could be configured along those lines :
javax/servlet/http/Cookie.setSecure(Z{1}):+HTTP_COOKIE_SECURE   -- If parameter is set to 1 add tag
javax/servlet/http/Cookie.setSecure(Z{0}):-HTTP_COOKIE_SECURE    -- If parameter is set to 0 remove tag

The detector would then trigger at the end of the analysis context to check if there are any objects of type X with a specific tag.

h3xstream:
@MaxNad I would not improve the configuration side for this. Some bug patterns will require the combination of multiple conditions, it will not supports those cases.

I would rather provide API to detectors that would follow specific events..

h3xstream:
@MaxNad 
Here is the first attempt.
https://github.com/h3xstream/find-sec-bugs/blob/feature/taint-extensions/plugin/src/main/java
/com/h3xstream/findsecbugs/password/HardcodedPasswordEqualsDetector.java

Life cycle:
1. Detector register as TaintFrameAdditionalVisitor
2. While the initial Taint Analysis run tags are added
3. Detector is called for each sink encounter

There is still unknown race conditions that make the local variables unstable. It will definitely not be integrated in the next release.

MaxNad:
I cannot seem to figure out how to look for missing calls using this.

If you take the Cookie example, you are trying to find cookies that do not have the setHttpOnly(true) call.
Using the current version of this detector, I believe this would look something like :
```
protected int getPriorityFromTaintFrame(TaintFrame fact, int offset) throws DataflowAnalysisException {
    if (!fact.getStackValue(offset).hasTag(Taint.Tag.HTTP_ONLY_COOKIE)) {
        return Priorities.NORMAL_PRIORITY;
    } else {
       return Priorities.IGNORE_PRIORITY;
    }
}

protected InjectionPoint getInjectionPoint(InvokeInstruction invoke, ConstantPoolGen cpg,
                                           InstructionHandle handle) {
    if(COOKIE_INIT_METHOD.matches(invoke,cpg)) {
        return new InjectionPoint(new int[]{0}, HTTPONLY_COOKIE_TYPE);
    }
    return InjectionPoint.NONE;
}

public void visitInvoke(InvokeInstruction instruction, ConstantPoolGen cpg, MethodGen methodGen, TaintFrame frameType) {
    try {
        if (SET_HTTP_METHOD.matches(instruction, cpg) && frameType.getValue(1).getIntegerConstantValue().intValue() == TRUE_INT_VALUE) {
            Taint value = frameType.getValue(0);
            value.addTag(Taint.Tag.HTTP_ONLY_COOKIE);
        }
    } catch (Exception e) {
        e.printStackTrace();
    }
}
```

However, _getPriorityFromTaintFrame_ will be called on the constructor and the HTTP_ONLY_COOKIE tag will be missing.

We could add the injection sink on the methods sending the cookies out to the client, but this would have to be modified every time a new library is added.

h3xstream:
@MaxNad Yup additional triggers will be need to be create.

>However, getPriorityFromTaintFrame will be called on the constructor and the HTTP_ONLY_COOKIE tag will be missing.
> We could add the injection sink on the methods sending the cookies out to the client, but this would have to be modified every time a new library is added.

## Thinking..

Basically, the InjectionDetector pattern is not corresponding to the cookie use case. The tagging part would be as-is. A different detector could iterate over the dataflow model later.

h3xstream:
I'm closing this PR. I will link eventually the new implementation of cookie verification.. (by the end of the year)

Issue #224, opened 2016-09-29
FP: sending local broadcasts via LocalBroadcastManager


luc-lynx:
Such piece of code 

```
LocalBroadcastManager.getInstance(getActivity()).sendBroadcast(intent);
```

causes a false positive alert "Broadcast intents could be received by a malicious application", but the code actually sends local broadcast that can't be sniffed by other applications.

MaxNad:
I do think the LocalBroadcastManager is safe to use.

https://developer.android.com/reference/android/support/v4/content/LocalBroadcastManager.html
    → "You know that the data you are broadcasting won't leave your app, so don't need to worry about leaking private data."

h3xstream:
This enhancement was fully implemented by @MaxNad.

luc-lynx:
Thanks!

PR #225, opened 2016-10-01
Issue #224 - Added an exception for the LocalBroadcastManager in the detector.

MaxNad:
I added a check in the BroadcastDetector to validate that the sendBroadcast call is not executed on the LocalBroadcastManager class. This should fix issue #224 .

Right now it could be fooled if someone calls its Activity LocalBroadcastManager and uses the global sendBroadcast method. I could change the check for something stricter like !"android/support/v4/content/LocalBroadcastManager".equals(getClassConstantOperand()) but this would have to be changed often since it contains the API version.

### DIFF
--- a/plugin-deps/src/main/java/android/content/LocalBroadcastManager.java
+++ b/plugin-deps/src/main/java/android/content/LocalBroadcastManager.java
@@ -1,0 +1,10 @@
+package android.content;
+
+public class LocalBroadcastManager {
+    public static LocalBroadcastManager getInstance(Context context) { return new LocalBroadcastManager(); }
+
+    //public void registerReceiver(BroadcastReceiver receiver, IntentFilter filter) {};
+    public boolean sendBroadcast(Intent intent) { return true; };
+    public void sendBroadcastSync(Intent intent) {};
+    public void unregisterReceiver(BroadcastReceiver receiver) {};
+}
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/android/BroadcastDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/android/BroadcastDetector.java
@@ -17,6 +17,7 @@
  */
 package com.h3xstream.findsecbugs.android;
 
+import com.h3xstream.findsecbugs.common.InterfaceUtils;
 import edu.umd.cs.findbugs.BugInstance;
 import edu.umd.cs.findbugs.BugReporter;
 import edu.umd.cs.findbugs.Priorities;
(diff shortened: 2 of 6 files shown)


plugin/src/main/java/com/h3xstream/findsecbugs/android/BroadcastDetector.java
```diff
@@ -37,11 +37,14 @@ public void sawOpcode(int seen) {
...
-                getNameConstantOperand().equals("sendOrderedBroadcast") ||
-                getNameConstantOperand().equals("sendOrderedBroadcastAsUser")
-                )) {
+                (
+                    getNameConstantOperand().equals("sendBroadcast") ||
+                    getNameConstantOperand().equals("sendBroadcastAsUser") ||
+                    getNameConstantOperand().equals("sendOrderedBroadcast") ||
+                    getNameConstantOperand().equals("sendOrderedBroadcastAsUser")
+                )
+                && !getClassConstantOperand().endsWith("LocalBroadcastManager")     // The LocalBroadcastManager object is safe. The broadcast doesn't leave the application scope.
```
h3xstream: I would use `InterfaceUtils.isSubtype()` has a main method to confirm that it is a LocalBroadcastManager.
The check `!getClassConstantOperand().endsWith("LocalBroadcastManager")`  can still be present. It will be useful to have fallback when the complete class hierarchy can be loaded (aka missing classes during the scan).

plugin/src/test/java/testcode/android/BroadcastIntentActivity.java
```diff
@@ -28,5 +29,12 @@ protected void onCreate(Bundle b) {
         sendOrderedBroadcast(i,null);
         sendOrderedBroadcast(i,null,null,null,0,null,null);
         sendOrderedBroadcastAsUser(i,null,null,null,null,0,null,null);
+
+        /* This call is safe.
+         *
+         * https://developer.android.com/reference/android/support/v4/content/LocalBroadcastManager.html
+         *      > "You know that the data you are broadcasting won't leave your app, so don't need to worry about leaking private data."
+         */
+        LocalBroadcastManager.getInstance(this).sendBroadcast(i);
```
h3xstream: Excellent doc. :+1:

plugin-deps/src/main/java/android/content/LocalBroadcastManager.java
```diff
@@ -0,0 +1,10 @@
+package android.content;
+
+public class LocalBroadcastManager {
```
h3xstream: I would add `android.support.v4.content.LocalBroadcastManager` type. It will be needed when an actual test on the class hierarchy is made. (See comment related to `InterfaceUtils.isSubtype()` below)

PR #234, opened 2016-10-25
Server Side Request Forgery (CWE 918) - Basic detector implementation

MaxNad:
Created a basic Scala Play detector for SSRF.

This detector check for Scala Play WebServer calls that accepts an unverified URL as parameter. This could be used by an attacker to bypass firewalls by making requests to internal services like mongodb.

This vulnerability also applies to Java applications but this detector doesn't check for it yet.


plugin/src/main/java/com/h3xstream/findsecbugs/scala/SSRFDetector.java
```diff
@@ -0,0 +1,34 @@
...
+import edu.umd.cs.findbugs.BugReporter;
+import edu.umd.cs.findbugs.Priorities;
+
+public class SSRFDetector extends BasicInjectionDetector {
+
+    private static final String SCALA_PLAY_SSRF_TYPE = "SCALA_PLAY_SSRF";
+
+    public SSRFDetector(BugReporter bugReporter) {
+        super(bugReporter);
+        loadConfiguredSinks("scala-play-ssrf.txt", SCALA_PLAY_SSRF_TYPE);
```
h3xstream: Excellent! This way we can add SSRF signatures with a different description (ie: for Java).

h3xstream:
Perfect !

PR #239, opened 2016-12-06
UNENCRYPTED_SERVER_SOCKET: use of java.net.ServerSocket

edrdo:
I added support for detecting the use of plain java.net.ServerSocket instances, similar to the one already in place for java.net.Socket, and suggesting the use of SSL server sockets (through SSLServerSocketFactory). 

The new UNENCRYPTED_SERVER_SOCKET vulnerability was relatively straightforward to integrate, it is similar to the one for UNENCRYPTED_SOCKET.
 
Please check if this is correct and can be integrated. The tests are passing, including the new one for UNENCRYPTED_SERVER_SOCKET.  (I just did not update the Japanese translation metadata). 

Best,
Eduardo Marques


h3xstream:
Everything is good ! Good description, good detector, good tests. 👍

edrdo:
Great :) 

Thanks!

PR #246, opened 2016-12-28
Adding ThreadLocalRandom detection

plr0man:
Adding detection of ThreadLocalRandom, insecure random number generator since Java 7.


h3xstream:
Everything is perfect. 👍
Thanks

PR #248, opened 2016-12-29
Adding overly permissive CORS policy detector

plr0man:
Detector reports overly permissive CORS policy that sets asterisk as the value of the Access-Control-Allow-Origin header, like in the following examples:
response.addHeader("Access-Control-Allow-Origin", "\*");
response.setHeader("Access-Control-Allow-Origin", "\*");

### DIFF
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/PermissiveCORSDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/PermissiveCORSDetector.java
@@ -81,8 +81,9 @@ public class PermissiveCORSDetector implements Detector {
 
                     LDC ldc = ByteCode.getPrevInstruction(location.getHandle().getPrev(), LDC.class);
                     if (ldc != null) {
-                        if ("Access-Control-Allow-Origin".equals(ldc.getValue(cpg)) &&
-                            "*".equals(ByteCode.getConstantLDC(location.getHandle().getPrev(), cpg, String.class))) {
+                        String headerValue = ByteCode.getConstantLDC(location.getHandle().getPrev(), cpg, String.class);
+                        if ("Access-Control-Allow-Origin".equalsIgnoreCase((String)ldc.getValue(cpg)) &&
+                            (headerValue.contains("*") || "null".equalsIgnoreCase(headerValue))) {
 
                             JavaClass clz = classContext.getJavaClass();
                             bugReporter.reportBug(new BugInstance(this, PERMISSIVE_CORS, Priorities.HIGH_PRIORITY)
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/PermissiveCORSDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/PermissiveCORSDetector.java
@@ -1,0 +1,103 @@
+/**
+ * Find Security Bugs
+ * Copyright (c) Philippe Arteau, All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3.0 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library.
+ */
+package com.h3xstream.findsecbugs;
+
+import com.h3xstream.findsecbugs.common.ByteCode;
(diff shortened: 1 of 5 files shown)


h3xstream:
Good one !

plugin/src/main/java/com/h3xstream/findsecbugs/PermissiveCORSDetector.java
```diff
@@ -0,0 +1,103 @@
...
+                INVOKEINTERFACE invoke = (INVOKEINTERFACE) inst;
+                String methodName = invoke.getMethodName(cpg);
+                String className = invoke.getClassName(cpg);
+
+                if (className.equals("javax.servlet.http.HttpServletResponse") &&
+                   (methodName.equals("addHeader") || methodName.equals("setHeader"))) {
+
+                    LDC ldc = ByteCode.getPrevInstruction(location.getHandle().getPrev(), LDC.class);
+                    if (ldc != null) {
+                        if ("Access-Control-Allow-Origin".equals(ldc.getValue(cpg)) &&
```
h3xstream: I would also test for the same header with different case (`access-control-allow-origin`). 
http://stackoverflow.com/a/5259004/89769

plugin/src/main/java/com/h3xstream/findsecbugs/PermissiveCORSDetector.java
```diff
@@ -0,0 +1,103 @@
...
+                            "*".equals(ByteCode.getConstantLDC(location.getHandle().getPrev(), cpg, String.class))) {
```
h3xstream: I would also test for wildcard use as sub-domain. (replace `equals` by `contains`)
http://blog.portswigger.net/2016/10/exploiting-cors-misconfigurations-for.html

Also the exact value "null" could be detected

plugin/src/main/java/com/h3xstream/findsecbugs/PermissiveCORSDetector.java
```diff
@@ -0,0 +1,103 @@
...
+
+            Instruction inst = location.getHandle().getInstruction();
+
+            if (inst instanceof INVOKEINTERFACE) {
+                INVOKEINTERFACE invoke = (INVOKEINTERFACE) inst;
+                String methodName = invoke.getMethodName(cpg);
+                String className = invoke.getClassName(cpg);
+
+                if (className.equals("javax.servlet.http.HttpServletResponse") &&
+                   (methodName.equals("addHeader") || methodName.equals("setHeader"))) {
```
h3xstream: Bonus : Typically website that need to use `Access-Control-Allow-Origin` will analyze the `Origin` header.
http://blog.portswigger.net/2016/10/exploiting-cors-misconfigurations-for.html

**This is not mandatory for this PR.** The description will have to be guideline of how to analyse properly Origin header. I would be interested to write this one.

ptamarit:
Maybe it would make sense to move this detector in a `cors` sub-package.

Not sure I'd have time to do it, but similar checks specific to Spring Framework could be added (see https://docs.spring.io/spring-security/site/docs/current/reference/html/cors.html and http://docs.spring.io/spring/docs/current/spring-framework-reference/html/cors.html).

plugin/src/main/java/com/h3xstream/findsecbugs/PermissiveCORSDetector.java
plr0man: Nice! Go ahead, I'm looking forward to seeing how you approach it.

plugin/src/main/java/com/h3xstream/findsecbugs/PermissiveCORSDetector.java
```diff
@@ -81,8 +81,9 @@ private void analyzeMethod(Method m, ClassContext classContext) throws CFGBuilde
 
                     LDC ldc = ByteCode.getPrevInstruction(location.getHandle().getPrev(), LDC.class);
                     if (ldc != null) {
-                        if ("Access-Control-Allow-Origin".equals(ldc.getValue(cpg)) &&
-                            "*".equals(ByteCode.getConstantLDC(location.getHandle().getPrev(), cpg, String.class))) {
+                        String headerValue = ByteCode.getConstantLDC(location.getHandle().getPrev(), cpg, String.class);
+                        if ("Access-Control-Allow-Origin".equalsIgnoreCase((String)ldc.getValue(cpg)) &&
```
plr0man: Fixed case sensitivity, header value wildcards and "null" value.

plugin/src/main/java/com/h3xstream/findsecbugs/PermissiveCORSDetector.java
```diff
@@ -81,8 +81,9 @@ private void analyzeMethod(Method m, ClassContext classContext) throws CFGBuilde
...
+                            (headerValue.contains("*") || "null".equalsIgnoreCase(headerValue))) {
```
plr0man: Fixed case sensitivity, header value wildcards and "null" value.

PR #250, opened 2016-12-31
Anonymous LDAP Bind detector

plr0man:
The proposed detector is reporting LDAP authentication method set to "none", like so:
`env.put(Context.SECURITY_AUTHENTICATION, "none");`
`DirContext ctx = new InitialDirContext(env);`


h3xstream:
Perfect!

PR #251, opened 2017-01-01
Persistent cookie detector

plr0man:
Proposed detector reports cookies set to expire in 1 year or more, like so:
`Cookie cookie = new Cookie("name", value);`
`cookie.setMaxAge(60*60*24*365);`


h3xstream:
Perfect!

PR #252, opened 2017-01-02
Add Url rewriting detector

plr0man:
Proposed detector reports URL rewriting methods. These methods determine whether the session ID needs to be included in the URL.

### DIFF
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/cookie/UrlRewritingDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/cookie/UrlRewritingDetector.java
@@ -1,0 +1,47 @@
+/**
+ * Find Security Bugs
+ * Copyright (c) Philippe Arteau, All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3.0 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


plugin/src/main/java/com/h3xstream/findsecbugs/cookie/UrlRewritingDetector.java
```diff
@@ -0,0 +1,47 @@
...
+    public UrlRewritingDetector(BugReporter bugReporter) {
+        this.bugReporter = bugReporter;
+    }
+
+    @Override
+    public void sawOpcode(int seen) {
+
+        if (seen == Constants.INVOKEINTERFACE && getClassConstantOperand().equals("javax/servlet/http/HttpServletResponse")
+                && (getNameConstantOperand().equals("encodeURL") || getNameConstantOperand().equals("encodeUrl") ||
+                getNameConstantOperand().equals("encodeRedirectURL") || getNameConstantOperand().equals("encodeRedirectUrl"))) {
```
h3xstream: I'm not sure to understand the objective of this detector.

Those method append potentially SESSIONID to the URL?

plr0man: Yes, if the browser does not support cookies those methods will return the URL with the session ID included. You'll find more information here:
[https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/http/HttpServletResponse.html#encodeURL(java.lang.String)](url)

h3xstream: Cool.

Issue #222, opened 2016-09-14
Changes to the Taint Analisys engine for Scala classes


MaxNad:
Hi @formanek & @h3xstream 

I was working on a Scala SQLi detector and I found a few problems with the taint analysis for Scala.
1. Scala uses some objects to represent Null values.
   For instance, the GETSTATIC scala/collection/immutable/Nil$.MODULE$ : Lscala/collection/immutable/Nil$; instruction is used to push an object representing a null list on the stack. The taint detector doesn't recognize this item as Null and this can lead to false positives.

I thought about adding a taint-config file listing such objects and handling them in a visitGETSTATIC method.
1. Scala has a method naming convention that is not recognized by the regex used in the taint analysis.

Ex. of unrecognized invoke instruction : anorm/package$SqlStringInterpolation$.SQL$extension(Lscala/StringContext;Lscala/collection/Seq;)Lanorm/SimpleSql;

Current Class Regex : ([a-z][a-z0-9]_\/)_[A-Z][a-zA-Z0-9\$]*
Could be change to : ([a-z][a-z0-9]_\/)_(package|[A-Z])[a-zA-Z0-9\$]*

Current Method Regex : (([a-zA-Z][a-zA-Z0-9]_)|(<init>))
Could be change to : (([a-zA-Z][a-zA-Z0-9]_(\$extension)?)|(<init>))

I just wanted to have your opinion on this before I start changing things into the taint analysis.

h3xstream:
@MaxNad I have no problem with this change of regex.

PR #254, opened 2017-01-03
SQL Injection (CWE-89) - Scala Slick & Scala Anorm injection detectors

MaxNad:
Added the Scala Slick and Scala Anorm API's to the sql injection detector.

I also added the changes to the Taint Analysis to increase its support of Scala classes (#222) since it was needed for this detector to work correctly.

This PR is linked to the vulnerable code samples in find-sec-bugs/find-sec-bugs-demos#6.


plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/TaintFrameModelingVisitor.java
```diff
@@ -190,6 +190,8 @@ public void visitSIPUSH(SIPUSH obj) {
 
     @Override
     public void visitGETSTATIC(GETSTATIC obj) {
+        // Scala uses some classes to represent null instances of objects
+        // If we find one of them, we will handle it as a Java Null
         if (obj.getLoadClassType(getCPG()).getSignature().equals("Lscala/collection/immutable/Nil$;")) {
```
h3xstream: What are the benefit of handling this correspondence? Are there any detector using this?
Nil could be added to the IMMUTABLE types .. https://github.com/find-sec-bugs/find-sec-bugs/blob/master/plugin/src/main/resources/taint-config/java-lang.txt

MaxNad: As of right now, the Scala Anorm injection-sink for SimpleSql.as would be using this.

When you use a StringContext without any parameters ( `s"test string"` ), the generated Byte code will call StringContext.s with a Nil object. 
`INVOKEVIRTUAL scala/Predef$.wrapRefArray ([Ljava/lang/Object;)Lscala/collection/mutable/WrappedArray;
    INVOKESPECIAL scala/StringContext.<init> (Lscala/collection/Seq;)V
    GETSTATIC scala/collection/immutable/Nil$.MODULE$ : Lscala/collection/immutable/Nil$;
    INVOKEVIRTUAL scala/StringContext.s (Lscala/collection/Seq;)Ljava/lang/String;`

I tried adding the Nil object to the IMMUTABLE types, but it doesn't seem to work. I think that since the Nil object was created using a GETSTATIC call, the taint-config is not used (I might be wrong there).

h3xstream: Ok. Perfect.
If this add to much complexity to TaintFrameModelVisitor, we will have to reconsider. At the moment, this one condition is reasonable and I don't see side effect.

PR #257, opened 2017-01-05
Fix false negatives in InsufficientKeySizeRsaDetector

plr0man:
1. Key length check up from 1024 to 2048 according to current standards and reflecting the existing bug description from messages.xml. Priority for key length between 1024 and 2048 set to low as opposed to normal for below 1024
2. Fixed the logic allowing for only one bug instance to be reported per file, now reporting all bug instances.

### DIFF
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/crypto/InsufficientKeySizeRsaDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/crypto/InsufficientKeySizeRsaDetector.java
@@ -74,7 +74,6 @@ public class InsufficientKeySizeRsaDetector implements Detector {
             Location location = i.next();
 
             Instruction inst = location.getHandle().getInstruction();
-//            ByteCode.printOpCode(inst, cpg);
 
             if (inst instanceof INVOKESTATIC) { //KeyPairGenerator.getInstance is called
                 INVOKESTATIC invoke = (INVOKESTATIC) inst;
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/h3xstream/findsecbugs/crypto/InsufficientKeySizeRsaDetector.java
```diff
@@ -118,22 +116,21 @@ private void analyzeMethod(Method m, ClassContext classContext) throws CFGBuilde
                         }
                     }
 
-                    if (n != null && n.intValue() < 1024) {
-                        initializeWeakKeyLength = true;
-                        locationWeakness = location;
+                    if (n != null && n.intValue() < 2048 && createRsaKeyGen) {
```
h3xstream: Good adjustment. Reporting the issue for 1024 bit key is perfect.
Even NIST which was one of the rare to allow 1024 bit for legacy key. This support has expired for key used in 2016+ (https://www.keylength.com/en/4/)

plugin/src/main/java/com/h3xstream/findsecbugs/crypto/InsufficientKeySizeRsaDetector.java
```diff
@@ -118,22 +116,21 @@ private void analyzeMethod(Method m, ClassContext classContext) throws CFGBuilde
...
-        if (createRsaKeyGen && initializeWeakKeyLength) {
-            JavaClass clz = classContext.getJavaClass();
-            bugReporter.reportBug(new BugInstance(this, RSA_KEY_SIZE_TYPE, Priorities.NORMAL_PRIORITY) //
-                    .addClass(clz)
-                    .addMethod(clz, m)
-                    .addSourceLine(classContext, m, locationWeakness));
-        }
+    private void addToReport(Method m, ClassContext classContext, Location locationWeakness, Number n){
+        JavaClass clz = classContext.getJavaClass();
+        int priority = (n.intValue() < 1024) ? Priorities.NORMAL_PRIORITY : Priorities.LOW_PRIORITY;
```
h3xstream: Excellent. **Priority** stands for **confidence** in FindBugs and this reflect perfectly the exploitability of each key size.

PR #259, opened 2017-01-06
Add Insecure SMTP SSL detector

plr0man:
Proposed detector reports SMTP SSL connections that don't verify the server certificate by default. This is equivalent to trusting all certificates. Affected APIs:
org.apache.commons.mail.Email
org.apache.commons.mail.HtmlEmail
org.apache.commons.mail.ImageHtmlEmail
org.apache.commons.mail.MultiPartEmail

A bug instance would be reported if the sslCheckServerIdentity property wasn't properly set in the below example:
`imageEmail.setSSLOnConnect(true);`
`imageEmail.setSSLCheckServerIdentity(true); //OK`


h3xstream:
Excellent detector!

PR #263, opened 2017-01-10
Add JavaBeans Property Injection detector

plr0man:
There are JavaBeans population methods designed for converting HTTP request parameters into a set of corresponding property setter calls on an arbitrary JavaBean. More info here:
https://commons.apache.org/proper/commons-beanutils/javadocs/v1.9.2/apidocs/org/apache/commons/beanutils/package-summary.html#conversion.beanutils
An attacker can leverage this functionality to access special bean properties like class.classLoader that will allow him to override system properties and potentially execute arbitrary code. Below is an example of a vulnerable code:

```java
MyBean bean = ...;
HashMap map = new HashMap();
Enumeration names = request.getParameterNames();
while (names.hasMoreElements()) {
    String name = (String) names.nextElement();
    map.put(name, request.getParameterValues(name));
}
BeanUtils.populate(bean, map); //reporting BeanUtils.populate() or BeanUtilsBean.populate()
```

Note: The proposed detector will report the above code based on the tainting of both HashMap.put() arguments coming from the request. However let me know if there is a way to customize the HashMap.put() taint so it only reports the "value" argument, which would be ideal in this situation.

### DIFF
--- a/plugin-deps/src/main/java/org/apache/commons/beanutils/BeanUtils.java
+++ b/plugin-deps/src/main/java/org/apache/commons/beanutils/BeanUtils.java
@@ -1,0 +1,10 @@
+package org.apache.commons.beanutils;
+
+import java.util.Map;
+
+public class BeanUtils {
+
+    public static void populate(Object bean, Map properties) {
+
+    }
+}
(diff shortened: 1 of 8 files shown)


h3xstream:
Please use for code sample
```
> ```java
> ```
```

h3xstream:
👍 Btw I really like this type of "niche" vulnerability. It reminds me of Struts RCE vulnerabilities.

PR #265, opened 2017-01-12
Add File Disclosure Injection detector

plr0man:
Proposed detector reports code utilizing ActionForward (Struts) or ModelAndView (Spring) to construct a server-side redirect path with user input. This may allow an attacker to view sensitive files or download application binaries.
Vulnerable code sample:
```java
... 
String returnURL = request.getParameter("returnURL");
return new ModelAndView(returnURL); 
```
or
```java
... 
String returnURL = request.getParameter("returnURL"); 
Return new ActionForward(returnURL);
```

### DIFF
--- a/plugin/src/main/resources/metadata/messages.xml
+++ b/plugin/src/main/resources/metadata/messages.xml
@@ -4824,4 +4824,73 @@ Avoid using user controlled values to populate Bean property names.
     </BugPattern>
     <BugCode abbrev="SECBPI">JavaBeans Property Injection</BugCode>
 
+
+    <!-- Struts and Spring File Disclosure -->
+    <Detector class="com.h3xstream.findsecbugs.injection.fileDisclosure.FileDisclosureDetector">
+        <Details>Detect Struts and Spring File Disclosure
+        </Details>
+    </Detector>
+    
+    <!-- Struts File Disclosure -->
+    <BugPattern type="STRUTS_FILE_DISCLOSURE">
+        <ShortDescription>Struts File Disclosure</ShortDescription>
+        <LongDescription>ActionForward populated with user controlled parameters</LongDescription>
+        <Details>
+            <![CDATA[
+<p>
+Constructing a server-side redirect path with user input could allow an attacker to download application binaries (including application classes or jar files) or view arbitrary files within protected directories.<br/>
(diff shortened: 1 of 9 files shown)


ptamarit:
I think that the following Spring snippet does the same risky redirects (returning a `String` instead of a `ModelAndView`:

```java
@GetMapping
public void index(...) {
    // ...
    return "redirect:" + request.getParameter("returnURL");
}
```

See:
* https://docs.spring.io/spring/docs/current/spring-framework-reference/html/mvc.html#mvc-redirecting-redirect-prefix
* http://stackoverflow.com/questions/19516719/difference-between-redirect-and-redirect-inside-modelandview

I'm afraid that writing a detector for this case would be trickier though...

plr0man:
Nice one Pablo! We'll have to pave the way for more of those. Do you want to add it to [Issues](https://github.com/find-sec-bugs/find-sec-bugs/issues) so we don't loose it?

ptamarit:
@plr0man: Done, I created #269.

plugin/src/main/resources/metadata/messages.xml
```diff
@@ -4824,4 +4824,73 @@ Avoid using user controlled values to populate Bean property names.
...
+
+    <!-- Spring File Disclosure -->
+    <BugPattern type="SPRING_FILE_DISCLOSURE">
+        <ShortDescription>Spring File Disclosure</ShortDescription>
+        <LongDescription>ModelAndView populated with user controlled parameters</LongDescription>
+        <Details>
+            <![CDATA[
+<p>
+Constructing a server-side redirect path with user input could allow an attacker to download application binaries (including application classes or jar files) or view arbitrary files within protected directories.<br/>
+An attacker may be able to forge a request parameter to match sensitive file locations. For example, requesting "http://example.com/?returnURL=WEB-INF/applicationContext.xml" would display the application's applicationContext.xml file. The attacker would be able to locate and download the applicationContext.xml referenced in the other configuration files, and even class files or jar files, obtaining sensitive information and launching other types of attacks.
```
php-coder: > For example, requesting "http://example.com/?returnURL=WEB-INF/applicationContext.xml" would display the application's applicationContext.xml file.

I believe that this example is wrong and user will always get 404 or 500 error. To do a server-side redirect you need to:
1) use `forward` method that isn't used by default. Hence, the parameter value should start from `forward:`
2) to get something in the WEB-INF dir, you need to add a slash to the beginning and only then this will be interpreted as a classpath resource.
3) now it still won't work because this string is passed to `ViewResolver` that most of the time configured to search in some directory and files that have particular suffix. Most likely you will be able to get JSP/HTML files only.
4) even if ViewResolver is configured to serve XML files, in modern Spring web apps (for example, Spring Boot) `applicationContext.xml` is absent at all because all the configuration is done in java classes

PR #266, opened 2017-01-14
Add Format String Manipulation Injection Detector

plr0man:
Proposed detector reports format string argument allowing user controlled parameters like so:
```java
...
+String format = "The customer: %s %s has the balance %4$." + userInput + "f";
+formatter.format(format, firstName, lastName, accountNo, balance);
```
Allowing user input to control format parameters could enable an attacker to leak information or cause exceptions to be thrown which could lead to application failure.

Affected APIs:
java/util/Formatter.format
java/io/PrintStream.printf
java/io/PrintStream.format
java/lang/String.format

### DIFF
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/injection/formatter/FormatStringManipulationDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/injection/formatter/FormatStringManipulationDetector.java
@@ -1,0 +1,44 @@
+/**
+ * Find Security Bugs
+ * Copyright (c) Philippe Arteau, All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3.0 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


h3xstream:
I'm not sure about the gain of integrating this one.
Format makes a lot of sense for C language where memory could be leaked.

I would prefer not giving confusion to user about a vulnerability that was probably never serious in Java. I might be wrong on this one..

plr0man:
This vulnerability is in fact not as serious in Java as it is in C. The description in message.xml covers possible consequences accordingly.
In case of Java you can modify the format string argument, such that an exception is thrown. If this exception is left uncaught, it may crash the application.
Alternatively, if sensitive information is used within the unused arguments, attackers may change the format string to reveal this information. The example code above lets the user specify the decimal points to which it shows the balance. If an attacker can specify the user input "2f %3$s %4$.2", the format string would be "The customer: %s %s has the balance %4$.2f %3$s %4$.2". This would then lead to the sensitive accountNo to be included within the resulting string.
Because other conditions have to be met before this flaw becomes a vulnerability, the priority was set to medium. The question here is how often those additional conditions are seen. Wouldn't causing the application to throw an uncaught exception a low priority alone?

plr0man:
I should've probably also included CERT's Secure Coding Standard for Java in the references:
[IDS06-J. Exclude unsanitized user input from format strings](https://www.securecoding.cert.org/confluence/display/java/IDS06-J.+Exclude+unsanitized+user+input+from+format+strings)

h3xstream:
Ok you convince me to integrate the detector.

PR #267, opened 2017-01-15
Add HTTP Parameter Pollution Injection Detector

plr0man:
Proposed detector reports HTTP Parameter Pollution (HPP), an attack that consist of injecting query string delimiters into other existing parameters. If a web application does not properly sanitize user input, a malicious user may compromise the logic of the application..
In the following example the attacker could provide a lang such as en&user_id=1, which would enable him to change the user_id at will.
```java
String lang = request.getParameter("lang");
+GetMethod get = new GetMethod("http://www.host.com");
+get.setQueryString("lang=" + lang + "&user_id=" + user_id);
+get.execute();
```

### DIFF
--- a/plugin-deps/src/main/java/org/apache/commons/httpclient/methods/GetMethod.java
+++ b/plugin-deps/src/main/java/org/apache/commons/httpclient/methods/GetMethod.java
@@ -1,0 +1,13 @@
+package org.apache.commons.httpclient.methods;
+
+public class GetMethod {
+
+    public GetMethod() {
+    }
+
+    public GetMethod(String uri) {
+    }
+
+    public void setQueryString(String queryString) {
+    }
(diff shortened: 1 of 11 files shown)


h3xstream:
The detector looks good.

I will probably renamed the tag HTTP_POLLUTION_SAFE to URL_ENCODED. It is likely that it will be use by other detector.

plr0man:
I would suggest naming it URL_PARAM_ENCODED instead of URL_ENCODED since the encoding required here is URL parameter specific. There are other methods for URL encoding that are less restrictive and used for whole URLs. Somebody might add one of those to the tag if it was named URL_ENCODED and affect the configuration of this particular detector.

h3xstream:
@plr0man ok

Issue #283, opened 2017-03-06
New rule: detect https connections with weak SSL / TLS protocol


xsun12:
Java by default uses SSL protocols or TLS 1.0 for outbound HTTPS connections. Both are insecure hence should be replaced by more secure TLS 1.2. Most developers are not careful enough to configure TLS protocol explicitly, hence there applications might be susceptible to man-in-the-middle attack.

Propose to add new rule to detect weak SSL / TLS protocols and recommend better practice.

I have 2 cases in mind:
1.  For applications using Apache HttpClient - detect the usage of [DefaultHttpClient ](http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/client/DefaultHttpClient.html)and recommend [SystemDefaultHttpClient](http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/client/SystemDefaultHttpClient.html)
1. For applications using [SSLContext ](http://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLContext.html)- detect the usage of "SSL" context and recommend "TLS" context

I can submit a pull request if this makes sense.

h3xstream:
This make perfect sense. I did some comments in the PR.

PR #285, opened 2017-03-10
Implement detector for weak SSL/TLS protocols

xsun12:
For issue #283, please kindly take a look.

### DIFF
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/crypto/WeakTLSDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/crypto/WeakTLSDetector.java
@@ -1,0 +1,69 @@
+/**
+ * Find Security Bugs
+ * Copyright (c) Philippe Arteau, All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3.0 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library.
+ */
+package com.h3xstream.findsecbugs.crypto;
+
+import edu.umd.cs.findbugs.BugInstance;
+import edu.umd.cs.findbugs.BugReporter;
+import edu.umd.cs.findbugs.OpcodeStack;
+import edu.umd.cs.findbugs.Priorities;
+import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
+import org.apache.bcel.Constants;
+
+public class WeakTLSDetector extends OpcodeStackDetector {
+
+    private static final String DEFAULT_HTTP_CLIENT = "DEFAULT_HTTP_CLIENT";
+    private static final String SSL_CONTEXT = "SSL_CONTEXT";
+
+    private BugReporter bugReporter;
+
+    public WeakTLSDetector(BugReporter bugReporter) {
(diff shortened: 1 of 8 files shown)


plugin/src/main/resources/metadata/messages.xml
```diff
@@ -981,6 +981,76 @@ uses. <b>PBKDF2</b> should be use to hash password for example.</p>
...
+<p>
+    <b>Vulnerable Code:</b><br/>
+<pre>
+    HttpClient client = new DefaultHttpClient();
+</pre>
+</p>
+<p>
+<p><b>Solution:</b><br/>
+
+Upgrade your implementation to use one of the recommended constructs and configure https.protocols JVM option to include TLSv1.2:</p>
```
h3xstream: Is `SystemDefaultHttpClient` a valid implementation ?
If so, it would be nice to have a valid sample solution.

plugin/src/main/java/com/h3xstream/findsecbugs/crypto/WeakTLSDetector.java
```diff
@@ -0,0 +1,69 @@
...
+            bugReporter.reportBug(new BugInstance(this, DEFAULT_HTTP_CLIENT, Priorities.NORMAL_PRIORITY)
+                    .addClass(this).addMethod(this).addSourceLine(this));
+        }
+        
+        if (seen == Constants.INVOKESTATIC 
+            && getClassConstantOperand().equals("javax/net/ssl/SSLContext") 
+            && getNameConstantOperand().equals("getInstance")
+            && getSigConstantOperand().equals("(Ljava/lang/String;)Ljavax/net/ssl/SSLContext;")) {
+          
+            System.out.println("SSLContext.getInstance(" + this.getSigConstantOperand() + ")");
```
h3xstream: Please comment this line. It will be noisy during scan.

plugin/src/main/java/com/h3xstream/findsecbugs/crypto/WeakTLSDetector.java
```diff
@@ -0,0 +1,69 @@
...
+package com.h3xstream.findsecbugs.crypto;
+
+import edu.umd.cs.findbugs.BugInstance;
+import edu.umd.cs.findbugs.BugReporter;
+import edu.umd.cs.findbugs.OpcodeStack;
+import edu.umd.cs.findbugs.Priorities;
+import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
+import org.apache.bcel.Constants;
+
+public class WeakTLSDetector extends OpcodeStackDetector {
```
h3xstream: The idea and the detector looks perfect.
The only missing component is a test case.

xsun12: I'm working on the test case for this the ApacheHttpClient library needs to be added as a test dependency - https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient.

I added it to plugin/test-dependencies.xml together with its own 3 dependencies, but it's not working. This is my first time working with Maven so I could use your help here.

h3xstream: FindSecBugs covers a lot of potential dependencies. The dependencies could also be from different versions. Instead of adding them as test dependencies.
We create stub class in this projet: https://github.com/find-sec-bugs/find-sec-bugs/tree/master/plugin-deps/src/main/java

You don't need to recreate all the methods .. Just the minimum need for compilation.

xsun12: Thanks for the pointer. Created a test case in WeakTLSDetectorTest.java.

plugin/src/main/resources/metadata/messages.xml
xsun12: yes SystemDefaultHttpClient is a valid implementation. Added sample solutions in the messages.

xsun12:
Addressed review comments. Please review again.

h3xstream:
Looks perfect!

PR #289, opened 2017-03-30
Spring Unvalidated Redirect Detector

johnhawes:
This adds a detector for unvalidated redirects in Spring MVC web applications.  The detector looks for a view prefix of "redirect:", which instructs the controller to perform a redirect to a specified URL.  The detector only looks at classes with "RequestMapping" annotations.  This should limit inspection to controllers and abstract classes that define endpoints.

### DIFF
--- a/plugin/src/test/java/testcode/spring/SpringUnvalidatedRedirectController.java
+++ b/plugin/src/test/java/testcode/spring/SpringUnvalidatedRedirectController.java
@@ -3,6 +3,7 @@ package testcode.spring;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.servlet.ModelAndView;
 
 @Controller
 public class SpringUnvalidatedRedirectController {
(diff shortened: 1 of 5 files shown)


h3xstream:
The log of travis-ci was unavailable when I last checked. Now it is available : https://travis-ci.org/find-sec-bugs/find-sec-bugs/builds/216934937

The build failure is related to missing.
```
[WARNING] Missing header in: /home/travis/build/find-sec-bugs/find-sec-bugs/plugin/src/main/java/com/h3xstream/findsecbugs/spring/SpringUnvalidatedRedirectDetector.java
[WARNING] Missing header in: /home/travis/build/find-sec-bugs/find-sec-bugs/plugin/src/test/java/com/h3xstream/findsecbugs/spring/SpringUnvalidatedRedirectDetectorTest.java
```

plugin/src/test/java/testcode/spring/SpringUnvalidatedRedirectController.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+
+@Controller
+public class SpringUnvalidatedRedirectController {
+
+    @RequestMapping("/redirect1")
+    public String redirect1(@RequestParam("url") String url) {
```
h3xstream: What if the return value is ModelAndView?

johnhawes: It should detect redirects for String and ModelAndView return values.  Essentially, the detector looks for string concatenation using the "redirect:" prefix in Spring controllers.  This could lead to false positives, but there should be limited false negatives.  I have updated the tests to confirm it will detect ModelAndViews and added LGPL license headers.

PR #305, opened 2017-05-15
Fixed copy-paste slip-up in Scala code example

ln2v:


MaxNad:
This is indeed a valid change. My bad. Thank you.

h3xstream:
Thanks for the fix.
No change is too small!

PR #306, opened 2017-05-31
Implement an unsafe jackson databind deserialization detector.

dbaxa:
Signed-off-by: David Black <dblack@atlassian.com>

### DIFF
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -169,6 +169,11 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.8.4</version>
+        </dependency>
 
         <!-- Unit Testing libraries -->
 
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -169,12 +169,6 @@
             <scope>test</scope>
         </dependency>
 
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-            <version>2.8.4</version>
-        </dependency>
-
         <!-- Unit Testing libraries -->
 
         <dependency>
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/serial/UnsafeJacksonDeserializationDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/serial/UnsafeJacksonDeserializationDetector.java
@@ -1,0 +1,139 @@
+/**
+ * Find Security Bugs
+ * Copyright (c) Philippe Arteau, All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3.0 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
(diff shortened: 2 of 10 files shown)


plugin/src/main/resources/metadata/messages.xml
```diff
@@ -4206,6 +4206,64 @@ Avoid deserializing object provided by remote users.
...
+    <Detector class="com.h3xstream.findsecbugs.serial.UnsafeJacksonDeserializationDetector">
+        <Details>Jackson databind Object deserialization detector</Details>
+    </Detector>
+
+    <BugPattern type="JACKSON_UNSAFE_DESERIALIZATION">
+        <ShortDescription>Unsafe Jackson deserialization configuration</ShortDescription>
+        <LongDescription>Unsafe Jackson deserialization configuration used in {1} {2} {3}</LongDescription>
+        <Details>
+            <![CDATA[
+<p>When the Jackson databind library is used incorrectly the deserialization of untrusted data can lead to remote code execution, if there is a class in classpath that allows the trigger of malicious operation.
```
h3xstream: Pretty cool. I did not know about this vulnerability.

plugin/pom.xml
```diff
@@ -169,6 +169,11 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.8.4</version>
```
h3xstream: This dependency is probably added to compile the sample.
To keep the project light and free of version conflict, we are using mock class and interface ([See plugin-dep](https://github.com/find-sec-bugs/find-sec-bugs/tree/master/plugin-deps/src/main/java)). The sample are not executed anyway.

In your case, you would need to create the stub classes for [com.fasterxml.jackson.annotation.JsonTypeInfo and 
com.fasterxml.jackson.databind.ObjectMapper](
https://github.com/find-sec-bugs/find-sec-bugs/pull/306/commits/a91cdf640dad328e167a7166007e4a536667be3f#diff-868763a9b824d007c96a6db59d31b4a2R3).

plugin/src/main/java/com/h3xstream/findsecbugs/serial/UnsafeJacksonDeserializationDetector.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private static final List<String> OBJECT_MAPPER_CLASSES = Arrays.asList(
+            "com.fasterxml.jackson.databind.ObjectMapper",
+            "org.codehaus.jackson.map.ObjectMapper");
+
+
+    @Override
+    public void visitClassContext(ClassContext classContext) {
+        JavaClass javaClass = classContext.getJavaClass();
+        if (OBJECT_MAPPER_CLASSES.contains(javaClass.getClassName())) {
```
h3xstream: Is it to avoid False Positive when scanning libraries (jackson or fasterxml)?

plugin/pom.xml
dbaxa: Okay will do.

plugin/src/main/java/com/h3xstream/findsecbugs/serial/UnsafeJacksonDeserializationDetector.java
dbaxa: Correct. This is to avoid marking the jackson classes themselves as false positives.

h3xstream:
Perfect !

PR #319, opened 2017-07-25
SSRF and LFI using RequestDispatcher and URLConnection

topolik:
Hi @h3xstream,

I extended detectors for vulnerabilities I met recently.

I reused `scala.SSRFDetector` class for the SSRF check even though it's not scala. I'm not sure if you would like to split it into 2 separate detectors - one for Scala and one for Java, or extract the `SSRFDetector` out of `scala` package and have just one class for both.

Thanks.


h3xstream:
Yup, I think it's fine in the SsrfDetector .. I will move it out of the `scala` package.
And it perfect to separate the LFI with request dispatcher because their are some additional/different concerns with the Spring redirects.

PR #331, opened 2017-08-18
Updated the CLI to 1.7.1

MaxNad:
This is a pretty small change.

I realized that the CLI was still using the Find Security Bugs 1.6.0 jar file.
I have updated the gradle build file and scripts.


h3xstream:
Thanks
Package added : https://github.com/find-sec-bugs/find-sec-bugs/releases/tag/version-1.7.1

PR #334, opened 2017-08-21
SpotBugs - 3.1.0-RC5

VinodAnandan:
https://github.com/spotbugs/spotbugs/releases


h3xstream:
What is the added value?

Issue #338, opened 2017-08-31
Enable assertions during building and/or using find-sec-bugs


javabeanz:
Find-sec-bugs has some assertions that guard correct execution of the code, but they are not enabled during testing, nor while running the plugin. This should be changed.

PR #339, opened 2017-08-31
Enable assertions during building and/or using find-sec-bugs #338

javabeanz:
part I : enabling assertions during testing


h3xstream:
Looks like a good plus .. There are a few asserts in the Taint Analysis.

PR #348, opened 2017-10-09
Renew Japanese messages.

orihalcon128:
I translated it into Japanese based on the latest version of messages.xml,
and fixed the language attribute value in bugs_ja.htm to "ja".

Tools:
- https://translate.google.co.jp/
- http://www.alc.co.jp/
- OmegaT (translation memory)
- Tomarigi (Japanese proofreading)

Reference:
https://vulncat.hpefod.com/


h3xstream:
Thank you for the contribution. I will have to trust you on this one. I know nothing about Japanese. 😕

Issue #138, opened 2015-12-11
More XXE coverage


h3xstream:
Some XXE parser to look for..

Built-in:
- [x] [TransformerFactory](https://github.com/find-sec-bugs/find-sec-bugs/blob/d4a0e0763d9cee71613b0fe4f00fa687a1084272/plugin/src/main/java/com/h3xstream/findsecbugs/xml/TransformerFactoryDetector.java) Added by @MaxNad
- [x] XMLReader
- [x] [XMLStreamReader/XMLInputFactory](https://github.com/find-sec-bugs/find-sec-bugs/blob/34e55f862ad7ba4cfc06a0f8b53bac6fa8354d3c/plugin/src/main/java/com/h3xstream/findsecbugs/xml/XmlStreamReaderDetector.java)
- [x] Unmarshaller
- [x] XPathExpression
- [x] SAXTransformerFactory (TransformerFactory is already convered)

More example here : https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Prevention_Cheat_Sheet

mcwww:
DOMParser must be added to this list as well, it's not detected at the moment by FindSecurityBugs

h3xstream:
@mcwww Is it oracle.xml.parser.v2.DOMParser (https://docs.oracle.com/cd/B14098_01/web.1012/b12024/oracle/xml/parser/v2/DOMParser.html)?

h3xstream:
Two issues were created to cover the API that could not be integrate in 1.8.0 (#394, #395)

ehats:
@h3xstream was support for Unmarshaller (or JaxbContext) ever implemented in the end? I couldn't find any references to it in the codebase.

h3xstream:
@ehats Good point.. I might have forgotten something.

h3xstream:
@ehats I have open #537

PR #349, opened 2017-10-12
Added some XXE Coverage for TransformerFactory

MaxNad:
I have created a detector to add some XXE coverage (related to issue #138).

This detector detects the use of the TransformerFactory class without the `FEATURE_SECURE_PROCESSING`, `ACCESS_EXTERNAL_DTD` or `ACCESS_EXTERNAL_STYLESHEET` flags.

I have noted during my tests that even if these flags are set, the TransformerFactory class is still vulnerable to Denial of Service attacks (like LoL Bombs). However, I have not found a way to completely remove entity processing in the classic TransformerFactory.

I have also discovered during my tests that the `XmlInputFactorySafe.parseXMLSafe2` method could also be vulnerable to XXE Denial or Service attacks. Maybe this should not be considered "safe" but should instead flag a lower severity vulnerability.

### DIFF
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/Taint.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/Taint.java
@@ -106,9 +106,6 @@ public class Taint {
         URL_ENCODED,
         PATH_TRAVERSAL_SAFE,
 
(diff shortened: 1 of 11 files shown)


h3xstream:
I'll review this by the end of the week.

h3xstream:
Looks good

Issue #356, opened 2017-11-07
New Rule: Detect Information Exposure through printStackTrace()


HansolChoe:
Hello.

Printing error messages to standard output may reveal security-sensitive information to users, and thus exposure of error messages would be vulnerable as reported by CWE-209 (https://cwe.mitre.org/data/deffinitions/209.html).

To detect possible exposures of error messages, I wrote a new detector that detects a method call to printStackTrace() as a bug, which possibly prints out sensitive system information to standard output. To reduce false alarms, I designed this detector focuses on the following 10 kinds of Exceptions which are also indicated by the CERT Oracle secure coding standard for Java:
* java.io.FileNotFoundException
* java.sql.SQLException
* java.net.BindException
* java.util.ConcurrentModificationException
* javax.naming.InsufficientResourcesException
* java.util.MissingResourceException
* java.util.jar.JarException
* java.security.acl.NotOwnerException
* java.lang.OutOfMemoryError
* java.lang.StackOverflowError

I implemented this detector, and now it is ready for the pull request. 
Please let me know if you like to send a pull-request for your review and merge. 

Thanks.

Hansol Choe

h3xstream:
The detector was merged in #360

PR #360, opened 2017-11-09
Add ErrorMessageExposureDetector #356

HansolChoe:
I submit this pull request as I created issue #356

Followings are changes I made:
- Create a new detector ErrorMessageExposureDetector at crypto
- Create a test class ErrorMessageExposureDetectorTest with 2 test cases
- Update findbugs.xml and messages.xml correspondingly

Thank you

### DIFF
--- a/plugin/src/main/java/com/h3xstream/findsecbugs/crypto/ErrorMessageExposureDetector.java
+++ b/plugin/src/main/java/com/h3xstream/findsecbugs/crypto/ErrorMessageExposureDetector.java
@@ -19,6 +19,7 @@ package com.h3xstream.findsecbugs.crypto;
 
 import edu.umd.cs.findbugs.BugInstance;
 import edu.umd.cs.findbugs.BugReporter;
+import edu.umd.cs.findbugs.OpcodeStack;
 import edu.umd.cs.findbugs.Priorities;
 import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
 import org.apache.bcel.Constants;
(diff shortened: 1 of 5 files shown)


plugin/src/main/resources/metadata/messages.xml
```diff
@@ -5669,4 +5669,35 @@ Sanitize user input before using it in HTTP parameters.
...
+<p>
+The sensitive information may be valuable information on its own (such as a password), or it may be useful for launching other, more deadly attacks. If an attack fails, an attacker may use error information provided by the server to launch another more focused attack. For example, an attempt to exploit a path traversal weakness (CWE-22) might yield the full pathname of the installed application. In turn, this could be used to select the proper number of ".." sequences to navigate to the targeted file. An attack using SQL injection (CWE-89) might not initially succeed, but an error message could reveal the malformed query, which would expose query logic and possibly even passwords or other sensitive information used within the query.
+</p>
+<p>
+    <b>Vulnerable Code:</b><br/>
+<pre>try {
+    /.../
+}
+catch (Exception e) {
+    e.printStackTrace();
```
h3xstream: This specific case does not seem to be vulnerable.

I would think of something like this:
```java
try {
  out = httpResponse.getOutputStream()

} catch (Exception e) {
  e.printStackTrace(out);
}
```
If the method `printStackTrace()` doesn't receive an outputStream, it will be output to the console. I don't see the exposure unless this is a console application expose over SSH/Telnet.

plugin/src/main/java/com/h3xstream/findsecbugs/crypto/ErrorMessageExposureDetector.java
```diff
@@ -0,0 +1,73 @@
...
+            case "java/sql/SQLException":
+            case "java/net/BindException":
+            case "java/util/ConcurrentModificationException":
+            case "javax/naming/InsufficientResourcesException":
+            case "java/util/MissingResourceException":
+            case "java/util/jar/JarException":
+            case "java/security/acl/NotOwnerException":
+            case "java/lang/OutOfMemoryError":
+            case "java/lang/StackOverflowError":
+                return true;
```
h3xstream: What if Throwable or Exception are catch.

```java
try {
  Connection conn = DriverManager.getConnection("jdbc:mysql://prod.company.com/production");
}
catch(Exception e) {
  e.printStackTrace(out);
}
```

I would added parent Exception. This will introduce some additional false positive but I think this is needed because catching Exception is very popular.

plugin/src/main/resources/metadata/messages.xml
h3xstream: Alternatively, the no argument signature could be set to LOW priority and the two other signatures (`printStackTrace(PrintStream s)`, `printStackTrace(PrintWriter s)`) could be set to normal because they are more likely to be seen by the user.

HansolChoe: Firstly, thank you for your code review and kind suggestions.

 I think the code you have written reveals the vulnerability better. It would be better that lines to be changed as you write:

```java
try {
  out = httpResponse.getOutputStream()

} catch (Exception e) {
  e.printStackTrace(out);
}
```

I also think it is good to distinguish `printStackTrace()` and to other signatures (`printStackTrace(PrintStream s)`, `printStackTrace(PrintWriter s)`).

I think I could fix that code lines and resubmit pull request.

h3xstream:
@HansolChoe Let me know when you think you'll have the time to do the changes.

plugin/src/main/java/com/h3xstream/findsecbugs/crypto/ErrorMessageExposureDetector.java
HansolChoe: I agree with your idea. Catching Exception or Throwable is very popular.

Adding Exception and Throwable could be a good solution.

Alternatively, how do you think about setting 10 vulnerable-to-print Exception classes calling `printStackTrace() ` case to NORMAL priority, and any other classes calling `printStackTrace()` case to LOW priority?

h3xstream: @HansolChoe I'm ok with either option. Reporting Low for every unknown exceptions will be slightly noisy.

HansolChoe:
Followings are changes I made:

- Update ErrorMessageExposureDetector
- Update ErrorMessageExposureDetectorTest with 3 test cases
- Update messages.xml

When 10 vulnerable-to-print classes and their parent classes(Throwable, Exception, or Error) call printStackTrace(), ErrorMessageExposureDetector alrams. 

If `printStackTrace(PrintStream s)` or `printStackTrace(PrintWriter w)` are called, ErrorMessageExposureDetector alarms with NORMAL priority. On the other hand, If  `printStackTrace()` is called, ErrorMessageExposureDetector alarms with LOW priority.

messages.xmlis modified with code you suggested.

h3xstream:
All good !
Thank you for your first contribution.

PR #373, opened 2018-03-22
OracleJDK9 for TravisCI

h3xstream:
Inspiration from : https://github.com/IDPF/epubcheck/blob/df885774abbe0b84fd7934c44ba0ddbfad7f8c99/.travis.yml


h3xstream:
This will need to be retest later.

PR #389, opened 2018-05-08
Make Java 9 build mandatory for Travis-CI

h3xstream:
At some point oracle JDK was simply not available on Travis-CI.
Now, I realise that it is running but some test are failing because of class removed/deprecated in Java 9.


h3xstream:
That's it was a minor glitch in the validation tests.

Issue #353, opened 2017-10-30
False positive XSS when using OWASP taglib 


gvasquez-waypoint:
Seems like the SonarQube plugin (or perhaps FindBugs itself) is not considering the use of the taglib version of the OWASP Encoder, as it is reporting a violation in the following excerpt:

`<br />MEDIANO PLAZO (${e:forHtmlContent(left)})<br />`

Considering:

`<%@taglib prefix="e" uri="https://www.owasp.org/index.php/OWASP_Java_Encoder_Project" %>`

SonarQube Version 6.5 (build 27846) 
FindBugs plugin: 3.6.0 installed (Analyze Java code with SpotBugs 3.1.0-RC6.)

Also posted at [SpotBugs](https://github.com/spotbugs/spotbugs/issues/485)

Ref: https://www.owasp.org/index.php/OWASP_Java_Encoder_Project#tab=Use_the_Java_Encoder_Project

h3xstream:
Quick Note : This is a know limitation. "${e:forHtmlContent(left)}" is a dynamic expression that is compiled at runtime. JSP expression are not analyzed by FSB or more precisely consider as UNKNOWN.

There is still some plan to whitelist some specific strings.

bradflood:
@h3xstream regarding the temporary fixes recommended above: the first option (`<e:forHtml value="${left}" />`) still shows as a violation findsecbugs-jsp:XSS_JSP_PRINT.  

I believe it is missed because the class (for example, `org.owasp.esapi.tags.EncodeForHTMLTag`) is not included in [owasp.txt](https://github.com/find-sec-bugs/find-sec-bugs/blob/master/plugin/src/main/resources/safe-encoders/owasp.txt)

h3xstream:
@bradflood Excellent pointer. I assumed it was on the safe list.

h3xstream:
The pattern `${e:forHtmlContent( [ascii name]  ) }` is now ignored. (No more FP)

PR #398, opened 2018-06-03
OWASP Taglib

bradflood:
@h3xstream this is initial PR for fixing false positives related to the OWASP Encoder Taglib (https://github.com/find-sec-bugs/find-sec-bugs/issues/353) .  Before I went too far, I wanted to validate that this was on the right track.

There were several stub classes to create, and the current PR simply validates the current false positive.  

Please review and let me know if you have any initial feedback. If all goes well, I will complete this PR within the next two weeks


h3xstream:
I will merge the current PR .. If you want to had the support for whitelist common expression. (#399) You can do so in a new PR.

PR #405, opened 2018-06-12
Created Kotlin version for HardcodedPasswordEqualsDetector class - fix #387

### LINKED ISSUE #387: Kotlin support for hardcode password with `Intrinsics.areEqual()`
This bug/new detector was suggested by @mario-areias.

With the following kotlin code..
```kotlin
fun thisIsATest(arg:String) {
    if("" == arg) {
        println("test1")
    }
    if("".equals(arg)) {
        println("test2")
    }
    if("" != arg) {
        println("test3")
    }
}
```
It generate the following Java pseudo code (decompiled with JD)
```java
  public static final void thisIsATest(@NotNull String arg)
  {
    Intrinsics.checkParameterIsNotNull(arg, "arg");
    String str;
    if (Intrinsics.areEqual("", arg))
    {
      str = "test1";System.out.println(str);
    }
    if ("".equals(arg))
    {
      str = "test2";System.out.println(str);
    }
    if ((Intrinsics.areEqual("", arg) ^ true))
    {
      str = "test3";System.out.println(str);
    }
  }
```

## The problem

While FSB look for `"superpassword1".equals(var)`, it doesn't look for `Intrinsics.areEqual()`

## Implementation

Creating a new detector inspired from [HardcodedPasswordEqualsDetector](https://github.com/find-sec-bugs/find-sec-bugs/blob/master/plugin/src/main/java/com/h3xstream/findsecbugs/password/HardcodedPasswordEqualsDetector.java) would be one option. [Some code](https://github.com/find-sec-bugs/find-sec-bugs/blob/master/plugin/src/main/java/com/h3xstream/findsecbugs/password/HardcodedPasswordEqualsDetector.java#L105) will need to be reuse, subclassing the detector could be done.

mario-areias:
As discussed on #387


plugin/src/test/java/com/h3xstream/findsecbugs/kotlin/KotlinHardcodedPasswordEqualsDetectorTest.java
```diff
@@ -0,0 +1,44 @@
+package com.h3xstream.findsecbugs.kotlin;
```
h3xstream: At the moment the build is failing because the license header is missing.

```
[INFO] Checking licenses...
[WARNING] Missing header in: /home/travis/build/find-sec-bugs/find-sec-bugs/plugin-deps/src/main/java/org/jetbrains/annotations/NotNull.java
[WARNING] Missing header in: /home/travis/build/find-sec-bugs/find-sec-bugs/plugin-deps/src/main/java/kotlin/Metadata.java
[WARNING] Missing header in: /home/travis/build/find-sec-bugs/find-sec-bugs/plugin-deps/src/main/java/kotlin/jvm/internal/Intrinsics.java
[WARNING] Missing header in: /home/travis/build/find-sec-bugs/find-sec-bugs/plugin/src/main/java/com/h3xstream/findsecbugs/password/AbstractHardcodedPasswordEqualsDetector.java
[WARNING] Missing header in: /home/travis/build/find-sec-bugs/find-sec-bugs/plugin/src/main/java/com/h3xstream/findsecbugs/kotlin/KotlinHardcodedPasswordEqualsDetector.java
[WARNING] Missing header in: /home/travis/build/find-sec-bugs/find-sec-bugs/plugin/src/test/java/com/h3xstream/findsecbugs/kotlin/KotlinHardcodedPasswordEqualsDetectorTest.java
```

 - This check is made for Test and Detector classes. You can copy paster the license header from any other class.

 - All stub class in plugin-deps can be whitelist to avoid this check --> https://github.com/find-sec-bugs/find-sec-bugs/blob/master/pom.xml#L56-L82. (You can add pattern to cover future classes `/java/kotlin/**`, `org/jetbrains/**`)

plugin/src/test/java/com/h3xstream/findsecbugs/kotlin/KotlinHardcodedPasswordEqualsDetectorTest.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+public class KotlinHardcodedPasswordEqualsDetectorTest extends BaseDetectorTest {
+
+    @Test
+    public void detectHardCodePasswordsWithEquals() throws Exception {
+        String[] files = {
+                getClassFilePath("bytecode_samples/kotlin_hardcoded_password_equals.jar"),
```
h3xstream: This is perfect. Alternatively, you can write the code in Java using Kotlin API to simplify the development process.

plugin/src/test/java/com/h3xstream/findsecbugs/kotlin/KotlinHardcodedPasswordEqualsDetectorTest.java
mario-areias: Sorry about that. I just ran the individual tests locally. Fixed the license headers now.

plugin/src/test/java/com/h3xstream/findsecbugs/kotlin/KotlinHardcodedPasswordEqualsDetectorTest.java
mario-areias: Good to know. I might try that in the next detectors.

h3xstream: Kotlin bytecode is really close to Java bytecode. There is less difference compared to let say, Scala.

PR #410, opened 2018-06-19
Added tests for Object deserialisation for kotlin code

mario-areias:
I added a jar for this one. In this case, I tried to add the decompiled version as a class, but the tests were not passing. It was finding another bug regarding having a redundant null check. I thought about fix it, but then it would be different from the actual decompiled code.

The slightly modified example is here (had to rename the class so it does not conflict with the existing class)

```java
package testcode.serial;

import java.io.IOException;
import java.io.InputStream;
import java.io.ObjectInputStream;
import kotlin.Metadata;
import kotlin.TypeCastException;
import kotlin.jvm.internal.Intrinsics;
import org.apache.commons.io.input.ClassLoaderObjectInputStream;
import org.jetbrains.annotations.NotNull;
import testcode.sqli.UserEntity;

@Metadata(mv={1, 1, 10}, bv={1, 0, 2}, k=1, d1={"\000\032\n\002\030\002\n\002\020\000\n\002\b\002\n\002\030\002\n\000\n\002\030\002\n\002\b\002\030\0002\0020\001B\005��\006\002\020\002J\020\020\003\032\0020\0042\006\020\005\032\0020\006H\007J\020\020\007\032\0020\0042\006\020\005\032\0020\006H\007��\006\b"}, d2={"Lorg/vulnerable/kotlin/ObjectDeserialisation;", "", "()V", "deserializeObject", "Lorg/vulnerable/kotlin/UserEntity;", "receivedFile", "Ljava/io/InputStream;", "deserializeObjectWithInheritance", "object-serialisation"})
public final class ObjectDeserializationKotlin {

    @NotNull
    public final UserEntity deserializeObject(@NotNull InputStream receivedFile)
            throws IOException, ClassNotFoundException {
        Intrinsics.checkParameterIsNotNull(receivedFile, "receivedFile");
        ObjectInputStream in = new ObjectInputStream(receivedFile);
        try {
            Object object = in.readObject();
            if (object == null) {
                throw new TypeCastException("null cannot be cast to non-null type testcode.sqli.UserEntity");
            }
            return (UserEntity) object;
        } finally {
            in.close();
        }
    }

    @NotNull
    public final UserEntity deserializeObjectWithInheritance(@NotNull InputStream receivedFile)
            throws IOException, ClassNotFoundException {
        Intrinsics.checkParameterIsNotNull(receivedFile, "receivedFile");
        ClassLoaderObjectInputStream in = new ClassLoaderObjectInputStream(getClass().getClassLoader(), receivedFile);
        try {
            Object object = in.readObject();
            if (object == null) {
                throw new TypeCastException("null cannot be cast to non-null type testcode.sqli.UserEntity");
            }
            return (UserEntity) object;
        } finally {
            in.close();
        }
    }
}
```


mario-areias:
Not sure why the coverage dropped. I just added more tests. :man_shrugging:

h3xstream:
@mario-areias I notice that that coveralls is having some issues with the coverage calculation.

Thanks for the added tests.

PR #416, opened 2018-07-17
Added KotlinHardcodePasswordInMap detector and created new module for…

JoshCunninghame:
- Added Kotlin test class module for Kotlin based detectors
- Added class path resolution for classes in jar files.
- Added KotlinHardcodePasswordInMap detector

### DIFF
--- a/findbugs-test-util/src/test/java/com/h3xstream/findbugs/test/service/ClassFileLocator.java
+++ b/findbugs-test-util/src/test/java/com/h3xstream/findbugs/test/service/ClassFileLocator.java
@@ -53,7 +53,7 @@ public class ClassFileLocator {
         assertNotNull(url, "No jsp file found for the path = " + path);
         return getFilenameFromUrl(url);
     }
-    
+
     public String getJarFilePath(String path) {
         ClassLoader cl = getClass().getClassLoader();
         URL url = cl.getResource(path);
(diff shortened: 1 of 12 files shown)


findbugs-test-util/src/test/java/com/h3xstream/findbugs/test/service/ClassFileLocator.java
```diff
@@ -63,12 +63,17 @@ public String getJarFilePath(String path) {
...
         String filename;
-		try {
-			filename = url.toURI().getPath();
-		} catch (final URISyntaxException e) {
-			fail("Failed to get file path = " + url, e);
-			return null;
-		}
+        try {
+            if (url.getPath().contains(".jar!")) {
+                String jarRegex = "findsecbugs-test-kotlin-[\\d\\.]{5,}-tests\\.jar!";
```
JoshCunninghame: Best effort to resolve files from the test-kotlin module.  Existing URI resolution resolved to null

findbugs-test-util/src/test/java/com/h3xstream/findbugs/test/service/ClassFileLocator.java
```diff
@@ -63,12 +69,18 @@ public String getJarFilePath(String path) {
...
     private String getFilenameFromUrl(URL url) {
         String filename;
-		try {
-			filename = url.toURI().getPath();
-		} catch (final URISyntaxException e) {
-			fail("Failed to get file path = " + url, e);
-			return null;
-		}
+        try {
+            if (TEST_JAR_MATCHER.matcher(url.getPath()).find()) {
```
JoshCunninghame: Attempt to replace in-jar file path with classpath.  Original url returned null.  Not sure if this approach is the best solution or if there is another mechanism to locate classes within dependency jars.

h3xstream: I barely remember writing this code. 🤣

It seems generic enough for a utility class. (some other spotbugs plugins are using this)

plugin/src/test/java/testcode/password/VariousMap.java
```diff
@@ -33,6 +33,10 @@ public void hardcode4() {
         init(map4);
     }
 
+    public void hardcode5() {
+        Map<String, String> map5 = new HashMap<>();
+        map5.putIfAbsent("password", "secret1234");
```
h3xstream: Good addition!

PR #419, opened 2018-07-22
Kotlin hardcode password equals

JoshCunninghame:
- Removed kotlin_hardcoded_password_equals.jar
- Updated KotlinHardcodedPasswordEqualsDetectorTest to use kotlin test jar


h3xstream:
Excellent! Thanks for taking the time to make the code cleaner. 👍

Issue #408, opened 2018-06-15
Possible bug in DeserializationGadgetDetectorTest


mario-areias:
It seems `DeserializationGadgetDetectorTest` is using a class that does not exist

https://github.com/find-sec-bugs/find-sec-bugs/blob/master/plugin/src/test/java/testcode/serial/commonscollections4/InvokerTransformer.java#L3

It looks to me that `Transformer` class lives on `org.apache.commons.collections4.Transformer` rather than `org.apache.commons.collections4.comparator.Transformer`. 

@h3xstream let me know if you agree with that or if I missed something. In case this is a bug I will send a PR in your way :smile:

h3xstream:
Yes. You are right.

PR #422, opened 2018-07-22
Kotlin deserialisation gadget

JoshCunninghame:
- Removed kotlin_deserialisation-gadget.jar
- Updated DeserializationGadgetDetectorTest to use kotlin test jar
- Fixed https://github.com/find-sec-bugs/find-sec-bugs/issues/408


plugin-deps/src/main/java/org/apache/commons/collections4/Transformer.java
```diff
@@ -1,4 +1,4 @@
-package org.apache.commons.collections4.comparators;
+package org.apache.commons.collections4;
```
h3xstream: Good catch!

PR #423, opened 2018-07-23
Added test coverage for command injection for kotlin string and func apis

JoshCunninghame:
- Added tests to exercise Kotlin functional and string APIs when mapping code injection points to sinks
- Problem when tracing String.filter and String.filterNot.  Looking for clarification.


findsecbugs-test-kotlin/src/test/kotlin/com/h3xstream/findsecbugs/command/IdentityFunctionCommandInjection.kt
```diff
@@ -0,0 +1,97 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library.
+ */
+package com.h3xstream.findsecbugs.command
+
+abstract class IdentityFunctionCommandInjection {
+
+    fun `tained input executed after always true filter`(tainted: String) {
```
JoshCunninghame: Currently failing test, looking for some advice on how to proceed to fix this.

Decompiled code looks as below.  Is this data flow something that can be tracked with spotbugs/findsecbugs?

If this is not something that we can track with findsecbugs I'll remove this test.

`public final void tained input executed after drop while always false(@NotNull String tainted) {`
`    Intrinsics.checkParameterIsNotNull(tainted, "tainted");String $receiver$iv = tainted;`
`    int $i$f$dropWhile;`
`    int i = 0;`
`    int index$iv;`
`    for (int j = ((CharSequence)$receiver$iv).length(); i < j; index$iv++) {`
`        char it = $receiver$iv.charAt(index$iv);`
`        int $i$a$1$dropWhile;`
`        it = $receiver$iv;Intrinsics.checkExpressionValueIsNotNull(it.substring(index$iv), "(this as.java.lang.String).substring(startIndex)");`
`    }`
`    String postFunctionTained = 0 == 0 ?`
`    tmp51_48 : "";Runtime.getRuntime().exec(postFunctionTained);`
`}`

h3xstream:
I have created this issue #428 to not forget to investigate this case.

PR #430, opened 2018-08-22
Added Kotlin support for CRLF detector

mario-areias:
The changes I have made were specifically to address CRLF detector issues. However, they will also be applied to all other Injection Detectors since they share the same code.

### DIFF
--- a/findsecbugs-test-kotlin/src/test/kotlin/com/h3xstream/findsecbugs/injection/Slf4jSample.kt
+++ b/findsecbugs-test-kotlin/src/test/kotlin/com/h3xstream/findsecbugs/injection/Slf4jSample.kt
@@ -1,0 +1,29 @@
+package com.h3xstream.findsecbugs.injection
+
+import org.slf4j.Logger
+
+class Slf4jSample {
+
+    fun slf4j(log: Logger, tainted: DataClass, tainted2: String) {
+        val safe = ""
+        //Unsafe
+        log.info(tainted.input)
+        log.info(tainted.input, safe)
+        log.info(safe, tainted.input)
+        log.info(safe, arrayOf<Any>(tainted.input!!))
+        log.info(safe, safe, tainted.input)
+        log.info(tainted2)
+        log.info(tainted2, safe)
+        log.info(safe, tainted2)
+        log.info(safe, arrayOf<Any>(tainted2))
+        log.info(safe, safe, tainted2)
+
+        // safe
+
+        log.info(safe, arrayOf<Any>(""))
+    }
+
+    inner class DataClass {
+        var input: String? = null
+    }
+}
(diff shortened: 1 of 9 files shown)


plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/TaintFrameModelingVisitor.java
```diff
@@ -539,30 +540,51 @@ private TaintMethodConfig getMethodConfig(InvokeInstruction obj) {
...
+    private TaintMethodConfig getConfigWithReplaceTags(TaintMethodConfig config, String className, String methodName) {
+
+        ObjectConfiguration objectConfiguration = new ObjectConfiguration(className, methodName);
+
+        if (!objectConfiguration.isAClassThatCanReplaceString()) {
+            return config;
+        }
+
+        /*
+          When Kotlin compiles the String replace method (when using regex values) Kotlin creates several instructions in its place.
```
mario-areias: This issue was a bit complicated. When Kotlin compiles instructions like `tainted.replace("[\r\n]+".toRegex(), "")` it generates many instructions in its place. One of the issues is that when using Regex values, the `replace` method being called is from the `Regex` class rather the one in `StringsKt` class.

I didn't manage to find a way to extract the value of the regex at the time the `Regex#replace` method was called, so I had to store some state when the `Regex#<init>` method is being called and use it when `Regex#replace` method was used.

I don't like this solution, but it works. More than happy to change it in case you know of a better solution (particularly one where store state is not necessary).

findsecbugs-test-kotlin/src/test/kotlin/com/h3xstream/findsecbugs/injection/Slf4jSample.kt
```diff
@@ -0,0 +1,29 @@
...
+
+class Slf4jSample {
+
+    fun slf4j(log: Logger, tainted: DataClass, tainted2: String) {
+        val safe = ""
+        //Unsafe
+        log.info(tainted.input)
+        log.info(tainted.input, safe)
+        log.info(safe, tainted.input)
+        log.info(safe, arrayOf<Any>(tainted.input!!))
```
mario-areias: The detectors think this line is safe. From what I understood after much debugging, is that Kotlin compiles that line in several different instructions (because of double `!`), including an if to see if the String is actually not null. Somehow the detector think this is a SAFE instruction when in reality it should be UNKNOWN. 

I didn't manage to find a solution for this one, though. So in the test I skip this line from the assertions. If you know a solution for this problem, let me know.

h3xstream: I have created an issue: https://github.com/find-sec-bugs/find-sec-bugs/issues/432

plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/TaintFrameModelingVisitor.java
h3xstream: Good analysis. I will look in depth at your fix later.

h3xstream:
I will keep this PR open .. otherwiste I will forget about #432

mario-areias:
No problem. Let me know if you think of a solution for this problem and I can give a crack in implementing it.

h3xstream:
I will do like #423 and integrate it first and look at the exception after.

PR #434, opened 2018-10-24
Fix sh script, so that cli can run on unix systems

Seraphiana:
We've discovered that findsecbugs 1.8 doesn't run on unix using the cli. Seems to be an issue with the sh file.

As such, I've rewritten the script and updated it to use the correct version.

FYI the docs for CLI support are referencing the wrong version: 1.46 and 1.44 in different places.

### DIFF
--- a/cli/findsecbugs.sh
+++ b/cli/findsecbugs.sh
@@ -1,4 +1,5 @@
+#!/bin/bash
 
-SOURCE="${BASH_SOURCE[0]}"
+cd $(dirname $(readlink `[[ $OSTYPE == linux* ]] && echo "-f"` $0))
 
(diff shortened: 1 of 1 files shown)


cli/findsecbugs.sh
```diff
@@ -1,4 +1,5 @@
+#!/bin/bash
 
-SOURCE="${BASH_SOURCE[0]}"
+cd $(dirname $(readlink `[[ $OSTYPE == linux* ]] && echo "-f"` $0))
```
h3xstream: `cd` will probably not work if a relative path is passed.

`./findsecbugs.sh someapp.jar`

Assuming the CLI folder is set in the **PATH**.

h3xstream:
The PR by @thypon has fix this entirely. #447
I will try to remember to post when a release is ready.

PR #444, opened 2018-12-12
Fix TravisCI for JDK version 10

h3xstream:
oraclejdk10 was changed to openjdk10 based on a recent error message on Travis-CI.

```
oraclejdk10 is deprecated. See https://www.oracle.com/technetwork/java/javase/eol-135779.html for more details. Consider using openjdk10 instead.
```


h3xstream:
It works

PR #447, opened 2019-01-07
Improve and generalize the CLI unix launcher

thypon:
- Decouple the launcher from the find-sec-bugs version
- Made it runnable from any directory
- Fix a bug with certain JDK classloaders not able to load wildcards
jars.

**This work has been sponsored by Doyensec LLC** [Doyensec](https://doyensec.com/)

### DIFF
--- a/cli/findsecbugs.sh
+++ b/cli/findsecbugs.sh
@@ -1,4 +1,20 @@
+#!/bin/bash
 
 SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
+  SOURCE="$(readlink "$SOURCE")"
+  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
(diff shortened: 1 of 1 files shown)


h3xstream:
That's awesome. I am going to test the script for Linux.

cli/findsecbugs.sh
```diff
@@ -1,4 +1,12 @@
...
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
+  SOURCE="$(readlink "$SOURCE")"
+  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
+FINDBUGS_PLUGIN="$(find $DIR/lib/findsecbugs-plugin-* | sort --version-sort | tail -n1)"
 
-java -cp "$SOURCE/lib/\*:" edu.umd.cs.findbugs.LaunchAppropriateUI -quiet -pluginList lib/findsecbugs-plugin-1.8.0.jar -include include.xml $@
+java -cp $(echo $DIR/lib/*.jar | tr ' ' ':') edu.umd.cs.findbugs.LaunchAppropriateUI -quiet -pluginList $FINDBUGS_PLUGIN -include include.xml $@
```
h3xstream: `-include include.xml $@` should be `-include $DIR/include.xml $@`

If Spotbugs can't load the filter configuration, it will skip it silently.
If the tool is placed into the PATH, it will scan the target jar, but without the security rules. 😲

h3xstream:
Sorry for the delay. I only found a small glitch. See the comment above.

thypon:
I included also another fix for handling superdirs with spaces in the names.

PR #449, opened 2019-01-30
Remove graph module

h3xstream:
This module was used for various experiments last year. It is however not mature and will not be maintained in the near future.
To avoid any confusion, I am removing it. Only the Git history will remember it.


topolik:
While I understand the reasons I'm sad seeing it goes. 

I played with it (generated 2GB database for the Liferay portal code) and looked promising. But didn't have the time to dig into it. :(

h3xstream:
@topolik Their is many things that the graph model was missing.
I stumble upon https://lgtm.com recently. It is far more advanced than what I intended to implement. I only played with it with small projects. It is free for open-source projects.

There is also https://www.shiftleft.io/ which is a similar solution that supports custom graph queries. There is a trial option that I have not tried yet.

PR #457, opened 2019-03-03
Added Entity Leak Detector

karanb192:


plugin/src/test/java/testcode/spring/SpringEntityLeakController.java
```diff
@@ -0,0 +1,24 @@
...
+
+@Controller
+public class SpringEntityLeakController {
+	@RequestMapping("/api1")
+	public SampleEntity api1(@RequestParam("url") String url) {
+		return new SampleEntity("entity1");
+	}
+
+	@RequestMapping("/api2")
+	public SampleEntityTwo api2(@RequestParam("url") String url) {
```
h3xstream: It would be nice to also cover List of Entity.
I know type erasure is a thing but it could be infer based on casting done inside the method.

plugin/src/main/java/com/h3xstream/findsecbugs/spring/SpringEntityLeakDetector.java
```diff
@@ -0,0 +1,134 @@
...
+			for (JavaClass subclass: javaClass.getSuperClasses()) {
+				annotations.addAll(getAnnotationList(subclass));
+			}
+		} catch (Exception e) {
+		}
+
+		return annotations;
+	}
+	private void analyzeMethod(Method m, ClassContext classContext) {
+		JavaClass clazz = classContext.getJavaClass();
```
h3xstream: `hasRequestMapping()` could be applied to only scan method with annotation rather than the complete  controller.
🤔
I am not sure .. We can always adjust if it generate to much false positive.
I will do some tests before the release.

h3xstream:
I will find a couple of spring web application to test the FP rate.

Did you test the detector on a large Spring web application ?

PR #465, opened 2019-03-27
Restructuring the sub-modules

h3xstream:
Why rename the folder
 - Make sure the folder name reflect the artifact-id.
 - Make it clear which submodule are use for test samples.

I am thinking about moving the `testcode/` package containing the Java samples to a dedicated submodule just like the kotlin module created by @mario-areias.
This would be done later. (Not in this release)


h3xstream:
Also I realize that the previous milestone, the `pom.xml` files were not update to `1.9.0-SNAPSHOT`. 🤦‍♂️

PR #478, opened 2019-05-04
Fix NPE when interface has spring mvc annotations

### LINKED ISSUE #477: SpringEntityLeakDetector throw s NPE
## Environment

| Component          | Version |
| ------------------ | ------- |
| Maven              | 3.6.0   |
| Java               | jdk-11.0.2  |
| SpotBugs           | 3.1.12   |
| FindSecBugs        | 1.9.0   |

## Problem
I got some problem when run spotbugs with find-sec-bugs plugin.
The issue is that SpringEntityLeakDetector throws NPE when some interface has `@RequestMapping` annotations.
For feign client I need to put spring MVC annotations on interface.

I think that problem is from `MethodGenFactory#analyze`.
That method returns null when `method.getCode() == null`.
So, (`MethodGen methodGen = classContext.getMethodGen(m);`) methodGen can be null.
NPE is thrown on `List<String> classesToInspect = new ArrayList<>(Arrays.asList(methodGen.getReturnType().toString()));`

It looks like to solve the issue by skipping when methodGen is null.

Is there any workaround for this problem?
Let me know please.

Thank you!

## Code (If needed)

```java
// FeignClient
public interface RemoteClient {
    @RequestMapping(path = "/hello/{name}", method = RequestMethod.GET)
    Hello hello(@PathVariable String name);
}
```

I made sample project to reproduce this issue.
https://github.com/bananayong/spring-entity-leak-detector-issue

bananayong:
This PR fix #477

Use Type instance directly, to prevent NPE.
`classContext.getMethodGen(m)` can return null, when interface is scanned.

---
Spring cloud openfeign client uses spring mvc annotation, but these client are not controller.
These classes should be skipped for spring detectors.


bananayong:
I have not yet. But, You're right.

Client and server can share very same interface for consistency.
I don't think so it is good practice, because generally interface name for feign client includes `XXXClient` or `RemoteXXXService`. But It can be.

What about SpringEntityLeakDetector check only annotated class(`@Controller`, `@RestController`) and find `@RequestMapping` annotated methods(including superclasses and interfaces)?

It's not perfect solution Because Spring supports meta-annotations. But, I think it is best efforts.
Please, let me know your idea.

Thank you!

h3xstream:
@bananayong I would simply remove this exception from the detector SpringEntityLeakDetector. There is something fishy if the client is reusing database classes as DTO.

bananayong:
You mean that annotated methods should be reported whether it is for feign client or not, Right?
I agree.

Actually, My purpose is to resolve NPE. If NPE is resolved then I can exclude `FeignClient` through findBugs excludeFilter.xml.

I will modify my PR.
Please forget feign client checking logic.

fsonntag:
@h3xstream Any plans on getting this merged and released? :)

h3xstream:
I will integrate this change as I see it add a test case for the empty method (from interface).

Note: I will push in a moment a PR that makes changes to this detector to support generic and include some other refactoring to better support mass assignment detection.

h3xstream:
The detector was refactored #496

PR #488, opened 2019-07-18
190430-taint-method-propagation

topolik:


topolik:
Closing, will resend. Looks like my master diverged from the upstream.

PR #530, opened 2019-10-29
Introduce a properties file to avoid repeating the versions

h3xstream:
Build script used to package 1.10.1


h3xstream:
fda21c40ed285670db4a19987f0a44ae960647c8 The second commit remove the unnecessary step of rerunning the tests in the `after_success` phase.

PR #551, opened 2020-04-24
Add example for sql_injection_spring_jdbc with annotation

Marx314:
This is an open discussion as the description can be also given for @Value without additional information.

It still can be vulnerable if the annotation use some SpEL voodoo but I'm not sure if how exactly.

Example of a "vulnerable" application : 
```java
public class ClientRepository {
    @Value("${jdbc.query.create.client}")
    private String reqCreateClient;  // INSERT INTO client (no_client, name, mail, date_inscrit) VALUES (?, ?, ?, ?);
    @Autowired
    private JdbcTemplate jdbcTemplate;
     
    public int createClient(Client client) {
        Assert.notNull(client, "entity cannot be null");
        try {
            return jdbcTemplate.update(reqCreateClient, ps -> {
                Assert.notNull(client.getId(), "embedded id clientId cannot be null");
                ps.setInt(1, client.getId());
                ps.setString(2, client.getName());
                ps.setString(3, client.getMail());
                ps.setTimestamp(4, client.getDateInscription());
            });
        } catch (DuplicateKeyException e) {
            return update(e);
        }
    }
}
```

### DIFF
--- a/findsecbugs-plugin/src/main/resources/metadata/messages.xml
+++ b/findsecbugs-plugin/src/main/resources/metadata/messages.xml
@@ -2350,11 +2350,24 @@ Bind variables in prepared statements can be used to easily mitigate the risk of
     <pre>JdbcTemplate jdbc = new JdbcTemplate();
 int count = jdbc.queryForObject("select count(*) from Users where name = '"+paramName+"'", Integer.class);
 </pre>
+    <pre>@Value("properties")
+private String sql;
+
+public function count() {
+    JdcbOperation jdbc = new JdcbOperation();
+    int count = jdbc.query(sql);
+}</pre>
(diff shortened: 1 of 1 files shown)


h3xstream:
The example above should be safe because "reqCreateClient" is loaded from a properties.

### (This is a false positive)
At the moment, we don't do extra analysis `@Value` from Spring. It would be nice I guess to add some heuristic to avoid common false positive.

### Generalizing the payload
From my basic research the values stored in `@Value` will not change at runtime once loaded. They seem to be scoped specifically to system properties and properties files. (https://www.baeldung.com/spring-value-annotation)

I guess we could consider those properties as safe without even analyzing the expression.

Issue #541, opened 2020-01-16
RuntimeException when processing static method


simon-greatrix:
## Environment

SpotBugs: 3.1.12
FindSecBugs: 1.10.1

Using SpotBugs GUI

## Problem

Analysis fails with this exception:
The following errors occurred during analysis:
```
  Exception while analyzing ThingUser.<init>()V
    java.lang.RuntimeException: Unable to call ThingProvider.get()LThingProvider;
      At com.h3xstream.findsecbugs.taintanalysis.TaintFrameModelingVisitor.visitInvoke(TaintFrameModelingVisitor.java:599)
      At com.h3xstream.findsecbugs.taintanalysis.TaintFrameModelingVisitor.visitINVOKESTATIC(TaintFrameModelingVisitor.java:385)
      At org.apache.bcel.generic.INVOKESTATIC.accept(INVOKESTATIC.java:87)
      At edu.umd.cs.findbugs.ba.AbstractFrameModelingVisitor.analyzeInstruction(AbstractFrameModelingVisitor.java:84)
      At com.h3xstream.findsecbugs.taintanalysis.TaintFrameModelingVisitor.analyzeInstruction(TaintFrameModelingVisitor.java:129)
      At com.h3xstream.findsecbugs.taintanalysis.TaintAnalysis.transferInstruction(TaintAnalysis.java:90)
      At com.h3xstream.findsecbugs.taintanalysis.TaintAnalysis.transferInstruction(TaintAnalysis.java:51)
      At edu.umd.cs.findbugs.ba.AbstractDataflowAnalysis.transfer(AbstractDataflowAnalysis.java:135)
      At edu.umd.cs.findbugs.ba.Dataflow.execute(Dataflow.java:376)
      At com.h3xstream.findsecbugs.taintanalysis.TaintDataflowEngine.analyze(TaintDataflowEngine.java:183)
      At com.h3xstream.findsecbugs.taintanalysis.TaintDataflowEngine.analyze(TaintDataflowEngine.java:56)
      At edu.umd.cs.findbugs.classfile.impl.AnalysisCache.analyzeMethod(AnalysisCache.java:369)
      At edu.umd.cs.findbugs.classfile.impl.AnalysisCache.getMethodAnalysis(AnalysisCache.java:322)
      At com.h3xstream.findsecbugs.injection.AbstractTaintDetector.getTaintDataFlow(AbstractTaintDetector.java:142)
[...]

PR #552, opened 2020-04-28
541  RuntimeException when processing static method

topolik:
Fixing #541 - RuntimeException when processing static method

Thanks.


findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/TaintFrameModelingVisitor.java
```diff
@@ -276,6 +276,14 @@ public void visitPUTSTATIC(PUTSTATIC obj) {
             Taint staticTaint = taintConfig.getStaticFieldTaint(fieldSig, null);
             Taint t = getFrame().getTopValue();
             t = Taint.merge(t, staticTaint);
+
+            // we are escaping a method context into a global class context
+            // method parameters and variables make no sense there
+            // 1. clear any method parameters
+            t.clearParameters();
+            // 2. clear method variable indexes
+            t.invalidateVariableIndex();
```
h3xstream, line 285: Thanks for the fix and the clear inline documentation.

Issue #558, opened 2020-06-19
False positive for unsafe comparison of hash that are susceptible to timing attack


titicalvados:
## Environment

Java 11, Maven, SonarQube 6.7.7 (LTS) with findsecbugs latest plugin 

## Problem

Rule findsecbugs:UNSAFE_HASH_EQUALS
Reference CWE-203: Information Exposure Through DiscrepancyKey
(https://cwe.mitre.org/data/definitions/203.html 2)

findsecbugs considers some strings are hash and report an unsafe comparison of hash that are susceptible to timing attack whereas it is some randomly generated strings (UUID.randomUUID().toString())

I can’t explain why the shareId string is considered as a hash.
Any idea ?
Regards

exemple of code generating the issue:
## Code (If needed)

```java
//if no share match shareId return exception
if (foundFile.getShares().stream().noneMatch(a -> a.getShareId().equals(shareId))) {
		throw new ShareUnknownException(String.format("Unknown share identifier: %s", shareId));
}
```
Screenshot :

![SonarQube](https://user-images.githubusercontent.com/50456999/85122996-c06c1980-b227-11ea-8c97-bc2128d4f0c0.png)

h3xstream:
The false positive occurs because the keyword SHA is associate to the hash algorithm.
https://github.com/find-sec-bugs/find-sec-bugs/blob/master/findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/password/HashUnsafeEqualsDetector.java#L64

I could add "shared" in whitelist for this detector as it is a common keyword.

h3xstream:
This issue was tagged [wontfix] by mistake.

h3xstream:
Fix in #565

PR #565, opened 2020-06-28
Add exception for the keyword share #558

h3xstream:
See #558


h3xstream:
Travis-CI need to be reconfigured.. I ran the tests locally.

Issue #570, opened 2020-07-10
Add support for sort with -V in findsecbugs.sh


Kbaek11:
I'm running FindSecBugs cli in an alpine container which is using a newer BusyBox version.
As of BusyBox [version 1.30.0](https://www.busybox.net/) the sort is called with 
```
sort -V
```
and running the current FindSecBugs cli release is failing with a busybox error due to the outdated sort call in the [findsecbugs.sh](https://github.com/find-sec-bugs/find-sec-bugs/blob/master/cli/findsecbugs.sh#L10)
```
sort --version-sort
```

PR #573, opened 2020-08-17
Change --version-list to -V

h3xstream:
Change --version-list to -V to support busybox version of the sort command #570


h3xstream:
Some testing will be needed before the release.

PR #576, opened 2020-08-18
Initial version of a Vert.x Sql Client detector

pmlopes:
Signed-off-by: Paulo Lopes <pmlopes@gmail.com>

This is a initial pull request to add a sql injection detector for vert.x sql client. The code passes the tests but the build is failing due to missing headers on the mock API.

I am wondering how could one solve the issue, as it seems that the remaining mocks do not have any headers, so I'm a bit lost here how to fix it.

### DIFF
--- a/findsecbugs-plugin/src/main/resources/metadata/messages.xml
+++ b/findsecbugs-plugin/src/main/resources/metadata/messages.xml
@@ -2507,6 +2507,60 @@ DB.withConnection { implicit c =>
     </BugPattern>
     <BugCode abbrev="SECSSQLA">SQL Injection (Scala - Anorm)</BugCode>
 
+    <!-- SQL Injection Vertx -->
+    <BugPattern type="SQL_INJECTION_VERTX">
+        <ShortDescription>Potential SQL Injection with Vert.x Sql Client</ShortDescription>
+        <LongDescription>This use of {3} can be vulnerable to SQL injection (with Vert.x Sql Client)</LongDescription>
+        <Details>
+            <![CDATA[
+<p>
+The input values included in SQL queries need to be passed in safely.
+Bind variables in prepared statements can be used to easily mitigate the risk of SQL injection.
+Vert.x Sql Client API provide a DSL to build query with Java code.
+</p>
+<p>
+    <b>Vulnerable Code:</b><br/>
+    <pre>
+SqlClient.query( "select * from Customer where id=" + inputId ).execute(ar -> ...);
+</pre>
+</p>
+<p>
+    <b>Solution (using Prepared Statements):</b><br/>
+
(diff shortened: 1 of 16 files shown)


findsecbugs-samples-java/src/test/java/testcode/sqli/VertxSqlClient.java
```diff
@@ -0,0 +1,33 @@
+package testcode.sqli;
+
+import io.vertx.sqlclient.SqlClient;
+import io.vertx.sqlclient.SqlConnection;
+
+public class VertxSqlClient {
+
+    public void injection1(SqlClient client, String injection) {
+        client.query(injection);
```
h3xstream, line 9: Example are minimalist but I think the vulnerability type for these API is well known.

findsecbugs-plugin/src/main/resources/metadata/messages.xml
```diff
@@ -2507,6 +2507,60 @@ DB.withConnection { implicit c =>
...
+    .onFailure(err -> ...);
+</pre>
+
+    <b>Solution (using OWASP Encoder):</b><br/>
+
+    <pre>
+import org.owasp.esapi.Encoder;
+
+SqlClient
+    .query( "select * from Customer where id=" + Encoder.encodeForSQL(inputId) )
```
h3xstream, line 2544: While FSB eliminates false positive when encoding functions are use, they do not represent the best solution.

For instance, in the code given - if `inputId` is a string - a malicious inputId could be `123 union select * from Users`. The string would not be sanitized because the encoding function assume that the value is placed between single quotes or double quotes. A `union` statement or any form of sub-queries could be introduced as long as no single quote is used. (Ref: [MysqlCodec from ESAPI](https://github.com/ESAPI/esapi-java-legacy/blob/d7bb3a537a76954242db94a6dd2cdb4a88a2856c/src/main/java/org/owasp/esapi/codecs/MySQLCodec.java#L177-L188))
```java
SqlClient
    .query( "select * from Customer where id=" + Encoder.encodeForSQL(inputId) )
```

Bottom line, I would simply remove this second solution. The rest is perfect.

findsecbugs-samples-java/src/test/java/testcode/sqli/VertxSqlClient.java
pmlopes: Yes the vert.x API is fluent so there's just 1 way to pass a sql statement to be executed, there are 2 classes where this happens:

* SqlClient - means a connection managed client where users run queries ad-hoc without worring to get a connection and returning it to the pool.
* SqlConnection - means that the user is being responsible to control the connection and return it to the pool.

PR #593, opened 2020-08-28
Change to GH Action

h3xstream:
- GH Action is no longer running the tests
- Avoid scanning samples


h3xstream:
It is working .. Only the SpotBugs Action failed to authenticate to push annotations.
https://github.com/h3xstream/find-sec-bugs/actions/runs/228938582

Issue #538, opened 2019-12-17
False positive spring jdbctemplate SQL Injection


dcherix:
## Environment

| Component          | Version |
| ------------------ | ------- |
| Maven              | ?????   |
| Gradle             | 6.0.1   |
| Java               | 8   |
| SpotBugs           | 3.1.12  |
| FindSecBugs        | 1.8.0   |

## Problem

[Find Security Bugs | Security | SQL_INJECTION_SPRING_JDBC] when using PreparedStatementSetter our BatchPreparedStatementSetter. Due to the use of PreparedStatement it's a false positive error.

## Code

```java
public class BugSample1 {
 private JdbcTemplate jdbcTemplate;
  
  public void someSqlCode(Object o){
     this.jdbcTemplate.update(substitutor.replace(DELETE_PERMISSION), ps -> {
            ps.setString(1, o.getType());
            ps.setLong(2, o.getIdentifier());
        });
}
}
```

h3xstream:
Is the value DELETE_PERMISSION a constant string (final static) ?

dcherix:
Yes it is

yahanvesh:
Any update on this?

h3xstream:
Not yet evaluate in depth .. mark so that I don't forget about it.

h3xstream:
Both classes are now supported.

PR #603, opened 2020-10-06
New samples with StringSubstitutor were added #538

h3xstream:


h3xstream:
Merge will have to be squash.

Issue #608, opened 2020-10-20
Restore Codecov integration


h3xstream:
## Description

The test coverage is no longer include in Pull Request. The configuration of Jacoco is likely the cause base on [the build output](https://travis-ci.org/github/find-sec-bugs/find-sec-bugs/jobs/733448918). 

````
[INFO] 
[INFO] --- jacoco-maven-plugin:0.8.2:report (report) @ findsecbugs-plugin ---
[INFO] Skipping JaCoCo execution due to missing execution data file.

[...]

0.31s$ bash <(curl -s https://codecov.io/bash)
  _____          _
 / ____|        | |
| |     ___   __| | ___  ___ _____   __
| |    / _ \ / _` |/ _ \/ __/ _ \ \ / /
| |___| (_) | (_| |  __/ (_| (_) \ V /
 \_____\___/ \__,_|\___|\___\___/ \_/
                              Bash-20200917-4e8f14b
==> Travis CI detected.
    project root: .
    Yaml not found, that's ok! Learn more at http://docs.codecov.io/docs/codecov-yaml
==> Running gcov in . (disable via -X gcov)
==> Python coveragepy not found
==> Searching for coverage reports in:
    + .
--> No coverage report found.
    Please visit http://docs.codecov.io/docs/supported-languages
```

h3xstream:
Fix in #609

PR #609, opened 2020-10-20
Attempt to export JaCoCo coverage file #608

h3xstream:
First attempt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ jdk:
 # Activate container based infra https://docs.travis-ci.com/user/workers/container-based-infrastructure/
 sudo: false
 install: mvn clean -B -V
-script: mvn clean test -Pjacoco -Ptravisci -B -V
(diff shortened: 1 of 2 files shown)


h3xstream:
This didn't work https://stackoverflow.com/a/57547478/89769

h3xstream:
Small explantation on why JaCoco stopped working.
Jacoco seems to add extra argument to the `maven-surefire-plugin` and because the travisci was overriding it later jacoco never attached to the tests.
```xml
                        <groupId>org.apache.maven.plugins</groupId>
                        <artifactId>maven-surefire-plugin</artifactId>
                        <version>2.20</version>
                        <configuration>
                            <argLine>-Xmx3000m -Dbcel.dontCache=true</argLine>
```

Full explantation : https://stackoverflow.com/questions/18107375/getting-skipping-jacoco-execution-due-to-missing-execution-data-file-upon-exec

Issue #611, opened 2020-10-21
Interface sinks not working?


Quiark:
## Environment

| Component          | Version |
| ------------------ | ------- |
| Maven              |  n/a   |
| Gradle             | 5.6.4   |
| Java               | 8   |
| SpotBugs           | 4.4.4   |
| FindSecBugs        | 1.11   |

## Problem
I'm trying to add SSRF sinks for Spring `RestTemplate` as well as `RestOperations` interface (into `findsecbugs-plugin/src/main/resources/injection-sinks/scala-play-ssrf.txt`). They look like this:

```
org/springframework/web/client/RestTemplate.delete(Ljava/net/URI;)V:0

org/springframework/web/client/RestOperations.delete(Ljava/net/URI;)V:0

...etc...
```

calls on the concrete class `RestTemplate` do get picked up but the ones on interface don't. 

Test cases:

```kotlin
    fun keraa(arg: URI) {
        val ops: RestOperations = client

        val qhuhu = ops.delete(arg)
    }

    fun direct(arg: URI) {
        val qhuhu = client.delete(arg)
    }

```

same code, decompiled:

```
  public void keraa(java.net.URI);
    Code:
       0: aload_1
       1: ldc           #202                // String arg
       3: invokestatic  #25                 // Method kotlin/jvm/internal/Intrinsics.checkParameterIsNotNull:(Ljava/lang/Object;Ljava/lang/String;)V
       6: aload_0
       7: getfield      #85                 // Field client:Lorg/springframework/web/client/RestTemplate;
      10: checkcast     #87                 // class org/springframework/web/client/RestOperations
      13: astore_2
      14: aload_2
      15: aload_1
      16: invokeinterface #205,  2          // InterfaceMethod org/springframework/web/client/RestOperations.delete:(Ljava/net/URI;)V
      21: getstatic     #211                // Field kotlin/Unit.INSTANCE:Lkotlin/Unit;
      24: astore_3
      25: return

  public void direct(java.net.URI);
    Code:
       0: aload_1
       1: ldc           #202                // String arg
       3: invokestatic  #25                 // Method kotlin/jvm/internal/Intrinsics.checkParameterIsNotNull:(Ljava/lang/Object;Ljava/lang/String;)V
       6: aload_0
       7: getfield      #85                 // Field client:Lorg/springframework/web/client/RestTemplate;
      10: aload_1
[...]

PR #618, opened 2020-11-30
Add Spring framework RestTemplate SSRF sinks

Quiark:
Extension functions currently don't work because of #611


h3xstream (changes requested):
Looks good! 👌 
The only thing missing is at least one test cases.

I see many stub classes but none are used by test cases (not test samples).

karinagee-cod (approved):

PR #620, opened 2021-01-27
Link to the SpotBugs plugin for IntelliJ

seanf:
The old [FindBugs-IDEA plugin](https://plugins.jetbrains.com/plugin/3847?pr=idea) doesn't work with current versions of IntelliJ.


h3xstream (approved):
Looks good.

Issue #601, opened 2020-10-05
Detect usage of Apache BeanUtils.copyProperties as dangerous


h3xstream:
## Description

BeanUtils is a library that is doing automatic mapping to Java object.
It can cause arm when the attack controls part of the list of properties being sets. BeanUtils does not blacklist properties like class, classloader or other objects that are likely to load arbitrary classes and possibly run code.

## Code

```java
import org.apache.commons.beanutils.BeanUtils;

public void addNewUser(Map<String,String> input) {
    BeanUtils.copyProperties(objToLoad, input);
}
```

Spring has apparently the similar class. The behavior will need to be investigate to see if it has any risk.

```java
import org.springframework.beans.BeanUtils;

public void addNewUser(Map<String,String> input) {
    BeanUtils.copyProperties(input, objToLoad);
}
```

https://stackoverflow.com/a/19763359/89769

Mehrapi:
I have installed the SpotBugs Plugins in my Intellij-IDEA, and opened the Apache BeanUtils repo locally. But the SpotBugs found no bugs in the BeanUtils file, but lots of Bugs in the other files.
I'm not sure what to do in this case.

Mehrapi:
Can I take this?

h3xstream:
For clarification..
I attempt to reproduce the vulnerability and could get any abuse scenario..

The current status is that a small need to be done to confirm the exploitability.

Once confirmed, this PR can be integrated #629

hasnainmakada-99:
Hey @h3xstream i would like to work on this issue 😃

h3xstream:
@hasnainmakada-99 In case you didn't see the comment in the PR .. We are looking to reproduce the vulnerability.

BeanUtils now has a class called SuppressPropertiesBeanIntrospector that blocks access to the class attribute.
https://github.com/apache/commons-beanutils/blob/master/src/main/java/org/apache/commons/beanutils2/SuppressPropertiesBeanIntrospector.java

# Severity

"class" is probably not the only attribute that could cause harm. There could be gadgets that lead to specific exploitation but the initial vector I had in mind was the RCE with Tomcat classloader (high severity).
Thinking about it, I think that [marcelel's PR](https://github.com/find-sec-bugs/find-sec-bugs/pull/629) has some value but we would need to lower the severity as there are no universal RCE with those API.
It is however nice to notify developers that some abuse could be done and class, where properties are "mass-assigned", could be targeted.

[...]

PR #629, opened 2021-04-20
Detect usage of Apache BeanUtils as dangerous #601

marcelel:


h3xstream:
Did you reproduce the security in the latest version ?

I don't want to integrate the new signature as it require some testing.
I had done some test before and the `class` property appears to be blocked in the latest version.

I will documented stuff I fund in this PR.

h3xstream:
## References for a security guard

### Release notes

> In version 1.9.2 now a special BeanIntrospector class was added which allows
> suppressing this property. Note that this BeanIntrospector is NOT enabled by
> default! Commons BeanUtils is a low-level library, and on this layer it cannot
> be decided whether access to a certain property is legal or not. Therefore,
> an application has to activate this suppressing BeanIntrospector explicitly.
> This can be done with the following lines of code:
> ```
> BeanUtilsBean bub = new BeanUtilsBean();
> bub.getPropertyUtils().addBeanIntrospector(
>     SuppressPropertiesBeanIntrospector.SUPPRESS_CLASS);
> ```

http://commons.apache.org/proper/commons-beanutils/javadocs/v1.9.4/RELEASE-NOTES.txt

### Jira 

> We had some internal discussions about this topic. Because BeanUtils is a very low-level library and may be used widely I am reluctant to build in a change which ignores the class property by default. This may break existing code.
> 
> BeanUtils 1.9 intorduced the possibility to customize bean introspection. It should be possible to write a custom bean introspector which ignores the class property. We can implement such an introspector and ship it with BeanUtils. In order to make it active, it has to be registered explicitly at a BeanUtilsBean instance or the central BeanUtils object.

https://issues.apache.org/jira/browse/BEANUTILS-463

### PR for the fix

Expected behavior taken from a test case..
```
    public void testAllowAccessToClassProperty() throws Exception {
        final BeanUtilsBean bub = new BeanUtilsBean();
        bub.getPropertyUtils().removeBeanIntrospector(SuppressPropertiesBeanIntrospector.SUPPRESS_CLASS);
        final AlphaBean bean = new AlphaBean();
        String result = bub.getProperty(bean, "class");
        assertEquals("Class property should have been accessed", "class org.apache.commons.beanutils2.AlphaBean", result);
    }
```

https://github.com/apache/commons-beanutils/pull/7

marcelel:
No, I didn't reproduce the security. I checked detection of usage `BeanUtils.copyProperties`. What do you suggest then?

h3xstream (approved):
This should have been reviewed and merged much sooner.

PR #633, opened 2021-05-26
Add new detector for NORMALIZATION_AFTER_VALIDATION

baloghadamsoftware:
On untrusted input strings applications must employ input filtering and
validation mechanisms based on the strings' character data. For example,
to avoid cross-site scripting (XSS) vulnerabilities the <script> tag
must be forbidden in the input. Unicode strings must also be normalized
the same string can have many different representations. However, this
normalization must happen before validation and not after.

This patch adds a detector for this bug.

PR #635, opened 2021-05-28
Add new detector for MODIFICATION_AFTER_VALIDATION

baloghadamsoftware:
A string must not be modified after validation because it may allow an
attacker to bypass validation using a tricky string which becomes
malicious after the modification. For example, a program may filter out
the <script> tags from HTML input to avoid cross-site scripting (XSS) and
other vulnerabilities. If non-character code points are deleted from the
input following validation, an attacker may pass the string "<scr"+
"\uFDEF"+"ipt>" so that the validation check fails to detect the <script>
tag, but the subsequent removal of the non-character code pont creates a
<script> tag in the input.

This patch adds a detector for this bug.

### DIFF
--- a/findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/ModificationAfterValidationDetector.java
+++ b/findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/ModificationAfterValidationDetector.java
@@ -1,0 +1,66 @@
+/**
+ * Find Security Bugs
+ * Copyright (c) Philippe Arteau, All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3.0 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library.
+ */
+package com.h3xstream.findsecbugs;
+
+import edu.umd.cs.findbugs.BugInstance;
+import edu.umd.cs.findbugs.BugReporter;
+import edu.umd.cs.findbugs.Priorities;
+import edu.umd.cs.findbugs.OpcodeStack;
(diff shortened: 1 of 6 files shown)


baloghadamsoftware:
Although our heuristics is quite rough, it seems that we do not get any false positives because of it. We tested this on several open-source projects and we did not get any new finding with this patch: [link](https://codechecker-demo.eastus.cloudapp.azure.com/java-static-analysis/reports?items-per-page=100&run=FindSecBugs%20on%20MATSim-Libs%20%28user%3A%20find-sec-bugs,%20branch%3A%20master%29&run=FindSecBugs%20on%20SpotBugs%20%28user%3A%20find-sec-bugs,%20branch%3A%20master%29&run=FindSecBugs%20on%20Jenkins%20%28user%3A%20find-sec-bugs,%20branch%3A%20master%29&run=FindSecBugs%20on%20Bt%20%28user%3A%20find-sec-bugs,%20branch%3A%20master%29&run=FindSecBugs%20on%20Ttorrent%20%28user%3A%20find-sec-bugs,%20branch%3A%20master%29&run=FindSecBugs%20on%20NanoHTTPD%20%28user%3A%20find-sec-bugs,%20branch%3A%20master%29&newcheck=FindSecBugs%20on%20MATSim-Libs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20ModifyBeforeValidation%29&newcheck=FindSecBugs%20on%20SpotBugs%20%28user%3A%20baloghadamsoftware,%20branch%3A%20ModifyBeforeValidation%29&newcheck=FindSecBugs%20on%20Jenkins%20%28user%3A%20baloghadamsoftware,%20branch%3A%20ModifyBeforeValidation%29&newcheck=FindSecBugs%20on%20Bt%20%28user%3A%20baloghadamsoftware,%20branch%3A%20ModifyBeforeValidation%29&newcheck=FindSecBugs%20on%20Ttorrent%20%28user%3A%20baloghadamsoftware,%20branch%3A%20ModifyBeforeValidation%29&newcheck=FindSecBugs%20on%20NanoHTTPD%20%28user%3A%20baloghadamsoftware,%20branch%3A%20ModifyBeforeValidation%29)

h3xstream:
I have to admit I was hesitant to merge this along with #633.
I am not sure about the value of such detector. Will it ever find something useful?
I guess we could say it covers a case documented by CERT. ✅ 

I have found a limited number of bugs where Normalizsation was the cause. Therefore I have very limited samples.

But I am ready to give it a try. I will merge **both** PRs. If the False Positive stay low, I have no issue with it.
(Which should be the case as normalization is not often used.)

Issue #324, opened 2017-08-07
Add Paths.get() as source for Path traversal


h3xstream:
Detect `Paths.get(userInput)`

PR #645, opened 2021-07-28
Add Paths.get(Uri) as source for Path traversal

deepsan:
This adds the Paths.get(Uri) variant as a follow-up to https://github.com/find-sec-bugs/find-sec-bugs/issues/324
> From the Path instance built, it can be use to load the file using other API (File, FileInputStream, etc..).

This also brings the nio detector in parity with the scala version of the same detector(scala-path-traversal-in.txt) which handles `Uri`s


h3xstream (approved):
Excellent contribution 👌

PR #655, opened 2021-09-24
New option to handle parameters of public methods of public classes as tainted data

baloghadamsoftware:
Libraries accessing sensitive data must be protected against malicious usage. Therefore, the parameters of the public methods in its public classes must be handles as tainted data. This patch implements this behavior by introducing a new option `findescbugs.taint.taintedpublicmethodparameters`.


baloghadamsoftware:
I am not sure if I put the test into the correct place, but I could not find any test for the options.

baloghadamsoftware:
Not needed.

PR #657, opened 2021-10-07
New detector for SEI-CERT rule SEC02-J: Do not base security checks on untrusted sources

baloghadamsoftware:
A new detector the checks whether the same non-final  method of a non-final
class that comes from a user was called before the privileged action and inside
it. Since neither the class, nor the method is final the user provided class
may override this method to behave in an unexpected, bypassing security checks.


baloghadamsoftware:
I checked this on real projects and there were no stability or performance problems, neither did I get any false positive. Thus we can say that it is safe.

baloghadamsoftware:
Actually, I am open to any suggestion. The only requirement is that we have to cover the SEC02-J SEI CERT rule.

baloghadamsoftware:
Closing in favor of spotbugs/spotbugs#1810.

Issue #679, opened 2022-04-22
SpringEntityLeakDetector crashes with array types


gtoison:
## Environment

| Component          | Version |
| ------------------ | ------- |
| SpotBugs           | 4.5.3.0+|
| FindSecBugs        | 1.12|

## Problem

SpringEntityLeakDetector does not seem to support the case when an argument is an array:

```
     [java] The following errors occurred during analysis:
     [java]   Exception analyzing licenses.adapters.inbound.rs.AssetResource using detector com.h3xstream.findsecbugs.spring.SpringEntityLeakDetector
     [java]     java.lang.IllegalArgumentException: Invalid class name [Lorg/springframework/web/multipart/MultipartFile
     [java]       At edu.umd.cs.findbugs.classfile.ClassDescriptor.<init>(ClassDescriptor.java:59)
     [java]       At edu.umd.cs.findbugs.classfile.DescriptorFactory.getClassDescriptor(DescriptorFactory.java:128)
     [java]       At edu.umd.cs.findbugs.AnalysisCacheToRepositoryAdapter.loadClass(AnalysisCacheToRepositoryAdapter.java:90)
     [java]       At org.apache.bcel.Repository.lookupClass(Repository.java:65)
     [java]       At com.h3xstream.findsecbugs.spring.SignatureParserWithGeneric.typeToJavaClass(SignatureParserWithGeneric.java:69)
     [java]       At com.h3xstream.findsecbugs.spring.SignatureParserWithGeneric.getArgumentsClasses(SignatureParserWithGeneric.java:54)
     [java]       At com.h3xstream.findsecbugs.spring.SpringEntityLeakDetector.analyzeMethod(SpringEntityLeakDetector.java:108)
     [java]       At com.h3xstream.findsecbugs.spring.SpringEntityLeakDetector.visitClassContext(SpringEntityLeakDetector.java:69)
     [java]       At edu.umd.cs.findbugs.DetectorToDetector2Adapter.visitClass(DetectorToDetector2Adapter.java:76)
     [java]       At edu.umd.cs.findbugs.FindBugs2.lambda$analyzeApplication$1(FindBugs2.java:1108)
     [java]       At java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
     [java]       At edu.umd.cs.findbugs.CurrentThreadExecutorService.execute(CurrentThreadExecutorService.java:86)
[...]

PR #680, opened 2022-04-22
fix: handle arrays in SignatureParserWithGeneric

gtoison:
Proposed fix for https://github.com/find-sec-bugs/find-sec-bugs/issues/679
Replace regex parsing of the method signature by `edu.umd.cs.findbugs.ba.generic.GenericUtilities` to handle the case when a type in the method signature is an array.
Updated the unit test to include a case when a paramete and the return type are arrays.

### DIFF
--- a/findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/spring/SignatureParserWithGeneric.java
+++ b/findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/spring/SignatureParserWithGeneric.java
@@ -17,15 +17,20 @@
  */
 package com.h3xstream.findsecbugs.spring;
 
+import edu.umd.cs.findbugs.ba.AnalysisContext;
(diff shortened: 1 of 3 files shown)


h3xstream (approved):
👍 Thanks for the clean bug fix

kev22257:
Hi, I was wondering if there was a planned date to release a new version of FSB that includes this fix up to the Maven central repository?

martmagi:
Would be nice to get a 1.13.0 release with this fix, as it's been plaguing our builds for a while (8 months to be exact).

kev22257:
Hey there, any update on version 1.13.0 and getting this fix included?  It would be awesome to be able to upgrade SpotBugs.

kev22257:
Greetings again, possible to get this into 1.13.0?  Is there a planned date to release a next version?

ctmay4:
Any word on the release of 1.13.0?

perwramdemark:
+1

PR #683, opened 2022-07-11
fix: added "cash account" to the safe words, not a SHA password

gtoison:
We have a false positive on UNSAFE_HASH_EQUALS for a variable containing `cashAccount`, it is flagged because the rule looks for `sha`
This PR adds `cashAccount` to the list of safe words


h3xstream (approved):
👍

Issue #575, opened 2020-08-17
Potential regression in Taint Analaysis on Java 8


h3xstream:
Taken from https://github.com/find-sec-bugs/find-sec-bugs/issues/332#issuecomment-674015645

## Problem

FSB on different version of JDK don't report the same issue. As an exemple, `BenchmarkTest00051.java` from OWASP Benchmark is not reporting an issue on Java 8 while it is report on Java 9+.

## Code (If needed)

```java
@Override
	public void doPost(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
		// some code
		response.setContentType("text/html;charset=UTF-8");
		

		org.owasp.benchmark.helpers.SeparateClassRequest scr = new org.owasp.benchmark.helpers.SeparateClassRequest( request );
		String param = scr.getTheValue("BenchmarkTest00051");

		
		String a1 = "";
		String a2 = "";
		String osName = System.getProperty("os.name");
        if (osName.indexOf("Windows") != -1) {
        	a1 = "cmd.exe";
        	a2 = "/c";
        } else {
        	a1 = "sh";
        	a2 = "-c";
        }
        String[] args = {a1, a2, "echo " + param};

		ProcessBuilder pb = new ProcessBuilder(args);
		
		try {
			Process p = pb.start();
			org.owasp.benchmark.helpers.Utils.printOSCommandResults(p, response);
		} catch (IOException e) {
			System.out.println("Problem executing cmdi - java.lang.ProcessBuilder(java.lang.String[]) Test Case");
            throw new ServletException(e);
		}
	}
```

h3xstream:
I did not mention it explicitly at first but this is due to the lack of support for `StringConcatFactory.makeConcatWithConstants`. This new method is introduce by the compiler. (From : https://github.com/find-sec-bugs/find-sec-bugs/issues/332#issuecomment-674015645)

PR #690, opened 2022-11-06
Adding workaround for JDK > 8 invokedynamic tainting

oxeye-gal:
Hi,

We [@oxeye](https://oxeye.io) created a possible workaround for this previously discussed issue:

https://github.com/find-sec-bugs/find-sec-bugs/issues/575

We are not 100% sure that this is the optimal solution for this issue, so we would love to get your feedback on this so we can address your observations.

We also noticed that you have set the target version for fixing this issue for the next version (1.13 scheduled for December), is it still the due date?

Thanks


oxeye-daniel:
@h3xstream - is there any update on this?

findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/taintanalysis/TaintConfig.java
```diff
@@ -216,6 +216,12 @@ private boolean isFieldType(String typeSignature) {
         return typeSignature != null && typeSignature.length() > 2 && typeSignature.charAt(0) != 'L';
     }
 
+    public int getMethodIdParamCount(String methodId) {
+        String signature = methodId.substring(methodId.indexOf("("), methodId.length());
+        SignatureParser p = new SignatureParser(signature);
+        return p.getNumParameters();
+    }
```
h3xstream, line 223: SignatureParser might not handle long and double parameters (which are double at runtime) but for `makeConcatWithConstants` it should be fine.

h3xstream (approved):
I didn't investigate deeply this PR but it should fix the missing method for StringConcatFactory / makeConcatWithConstants from Java 9.

h3xstream:
Might fix #701

jbindel:
It seems to help greatly for JDK11 bytecodes  with the new option turned on. One thing that does not get fixed is taint propagation for `makeConcatWithConstants` when concatenating a `long` or `double` with strings, which I believe is what you noted.

`"something" + 1000L` is tainted while `"something" + Long.toString(1000L)` is clean.

jbindel:
I have some changes building on this pull request to make `TaintConfig` work when concatenating `long` and `double` values, and I'll make a pull request for them soon.

PR #691, opened 2022-11-06
Verbose source line locations report

oxeye-gal:
Hi,

In some cases it is useful to get all the byte code offsets find-sec-bugs aware of related to a single source code line, for example when one would like to get context to a specific location in a line.

We added a config flag that once set will skip the code that responsible for keeping only one SourceLineAnnotation per line.


h3xstream (approved):
Looks ok!

PR #712, opened 2023-08-18
taint-config files java-lang.txt and scala.txt propagate taint from character types

jbindel:
For Issue#711 the taint on StringBuilder and StringBuffer from character types is propagated, and Character class is no longer considered SAFE.

Because libraries may turn strings into character arrays and then rebuild the string, we cannot assume that adding characters individually to a string will be untainted.

```
for (char c : unsafe.toCharArray()) {
    // Maybe do other stuff but not necessarily remove unsafe sequences.
    buf.append(c);
}

// Later may build a stream from the characters that were in the unsafe string.
InputStream is = new FileInputStream("Hello" + buf);
```

This pull request changes the taint configuration for both the java-lang and scala classes that deal with individual characters.


h3xstream (approved):
I didn't get the time to look at the changes in depth.
Thanks for contributing. I will try to do a release ASAP.

PR #713, opened 2023-08-18
Updates to handle string-building taint with invokedynamic concatenation in JDK > 8

jbindel:
Pull Request https://github.com/find-sec-bugs/find-sec-bugs/pull/690 adds handling of JDK > 8 string concatenation with invokedynamic and `makeConcatWithConstants`.  There are some cases it does not handle such as concat of long or double values.  This pull request address the parameter offsets affected by long and double, which take two spots on the stack.

It also fixes issues where taint on a stack object is modified when we do not want that.

In order to test this change, we need some sample classes built with a JDK > 8, and using JDK11 on the current `findsecbugs-samples-java` introduces errors in the unit tests so there is a build submodule that builds specific classes to a JDK 11 target.

### DIFF
--- a/findsecbugs-plugin/pom.xml
+++ b/findsecbugs-plugin/pom.xml
@@ -148,6 +148,12 @@
         </dependency>
         <dependency>
             <groupId>com.h3xstream.findsecbugs</groupId>
+            <artifactId>findsecbugs-samples-java11</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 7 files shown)


h3xstream (approved):
Thanks for the contributions
I will do some Java 11 specific tests before the release.

jbindel:
To enable handling of Java 9+ string concatenation, we have to set the system property `findsecbugs.taint.workaroundvisitinvokedynamic` to `true`.  It may be worthwhile to enable this behavior by default as most people may not discover the setting, and Java 8 is becoming less common.

Issue #723, opened 2024-01-24
Java 21 Support


Jeeppler:
## Description

Support [Java JDK 21](https://jdk.java.net/21/). The release of Java 21 is the next long term support release.

I am willing to help implement it, but I would like to have some directions on how to get started.

haerter-tss:
Any updates on this issue? When are you planning to support java 21?

gtoison:
Have you encountered issues with Java 21?
SpotBugs is regularly tested with JDK 21 and it runs fine, I suppose that the find-sec-bugs plugin should work for the most part

Jeeppler:
SpotBugs does support JDK 21, but FindSecBugs does use an old version of SpotBugs. In essence, FindSecBug needs a new release to support Java 21.

gtoison:
I think it's normally the other way around: SpotBugs runs and uses FindSecBugs as a plugin
So what matters is the SpotBugs version you're using, plugins such as FindSecBugs will not override that version

Or maybe there's a way to run FindSecBugs standalone?

Jeeppler:
Yes, the problem is with the standalone version.

If you go to the latest [FindSecurityBugs release 1.12.0](https://github.com/find-sec-bugs/find-sec-bugs/releases/tag/version-1.12.0) and download the `findsecbugs-cli-1.12.0.zip` you will end up with the following structure:

~~~
user@pc:~/findsecbugs-cli-1.12.0$ tree
.
├── findsecbugs.bat
├── findsecbugs.sh
├── include.xml
└── lib
    ├── asm-9.2.jar
    ├── asm-analysis-9.2.jar
    ├── asm-commons-9.2.jar
    ├── asm-tree-9.2.jar
    ├── asm-util-9.2.jar
    ├── bcel-6.5.0.jar
    ├── commons-lang3-3.12.0.jar
    ├── commons-text-1.9.jar
    ├── dom4j-2.1.3.jar
    ├── findsecbugs-plugin-1.12.0.jar
    ├── gson-2.9.0.jar
    ├── jaxen-1.2.0.jar
    ├── jcip-annotations-1.0.jar
    ├── jsr305-3.0.2.jar
    ├── Saxon-HE-10.6.jar
    ├── slf4j-api-1.8.0-beta4.jar
    ├── spotbugs-4.6.0.jar
    └── spotbugs-annotations-4.6.0.jar

1 directory, 21 files
~~~

[...]

PR #725, opened 2024-02-21
Upgrade SpotBugs to 4.8.3

gtoison:
This should bring compatibility to JDK 21 (see #723)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -199,7 +199,7 @@
                     <plugin>
                         <groupId>com.github.spotbugs</groupId>
                         <artifactId>spotbugs-maven-plugin</artifactId>
-                        <version>4.8.3</version>
+                        <version>4.8.3.1</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -199,7 +199,7 @@
                     <plugin>
                         <groupId>com.github.spotbugs</groupId>
                         <artifactId>spotbugs-maven-plugin</artifactId>
-                        <version>4.5.3</version>
+                        <version>4.8.3</version>
```
pzygielo, line 202: Not sure how this works, but there is no such version in central.

gtoison: You're right, thanks for spotting this!
4.5.3 does not exist either, but the latest is 4.8.3.1

Jeeppler:
Is it possible to have a new release afterwards?

PR #729, opened 2024-02-28
Upgrade the build to run on JDK 11

gtoison:
The build currently fails because it runs on JDK 8 but the findsecbugs-samples-java11 module requires java 11

### DIFF
--- a/.github/workflows/spotbugs.yml
+++ b/.github/workflows/spotbugs.yml
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 1.8
+    - name: Set up JDK 11
       uses: actions/setup-java@v1
       with:
-        java-version: 1.8
+        java-version: 11
     - uses: actions/cache@v1
       with:
         path: ~/.m2/repository


.github/workflows/spotbugs.yml
```diff
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 1.8
+    - name: Set up JDK 11
```
pzygielo, line 12: Maybe version could be dropped from the name? Then - with the future change to come, there will be no need to keep them in sync.

gtoison: I'm used to see the JDK version in the build steps because when doing matrix builds (running the build on different JDK) it helps quickly seeing what's running. But I agree that we could remove it as well, I suppose it's matter of preference

pzygielo: With the matrix build, that is not the case here, the matrix element could be used to build names. Here the literal `1.8`/`11` is used (twice).

The matrix with one JDK would do as well.

Currently it's the only step in the job and there is no different JDK involved so far. Hence I proposed to consider dropping version from the name.

h3xstream: I am ok with being more explicit on the label name.

PR #734, opened 2024-04-30
Add support for jakarta

cheeghi:
This PR adds support for  `jakarta.*` for detectors that work with the old `javax.*` classes (Cookie, HttpServlet, Entity, etc.).
Unit tests have been added, except for jsp detectors that I think would require setting up a different jsp compilation (with a newer compiler).


findsecbugs-plugin/src/test/java/com/h3xstream/findsecbugs/PermissiveCORSDetectorTest.java
```diff
@@ -93,4 +93,23 @@ public void detectPermissiveCORS() throws Exception {
                         .build()
         );
     }
+
+    @Test
+    public void detectJakartaPermissiveCORS() throws Exception {
```
h3xstream, line 98: Excellent contribution!

h3xstream (approved):
Thanks! I heard about Jakarta but I didn't notice new package would be used.

PR #746, opened 2024-10-20
Local build fails after #740

topolik:
The PR #740 renamed ant `tasks` to maven `target` that requires `maven-antrun-plugin:3.0.0` or higher. 

My local env build fails, mvn doesn't download the latest dependency. Requiring the version manually fixes the problem.

### DIFF
--- a/findsecbugs-plugin/pom.xml
+++ b/findsecbugs-plugin/pom.xml
@@ -39,7 +39,9 @@
             </plugin>
             <!-- Moving findbugs config to root -->
             <plugin>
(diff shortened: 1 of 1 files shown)


jbindel:
For example, a standard find-sec-bugs build on Ubuntu 22.04 LTS with the Maven version from the Canonical repositories gives me `maven-antrun-plugin:1.3`, which won't work because the `target` element was added in version 1.5, so adding the explicit version for 3.1.0 (currently the latest is a good improvement.

h3xstream (approved):
Thanks.

I didn't had the opportunity to test it yet.

PR #750, opened 2024-12-10
Misinterpreting invokedynamic causes "missing class" warning

### LINKED ISSUE #332: invokedynamic from lambdas produces wrong missing class warnings
A lambda in a line like `fieldAccessors.entrySet().forEach(x -> sb.append(x.getKey()).append(" = ").append(x.getValue()).append("\n"));` is compiled to an `invokedynamic` bytecode instruction.

In `com.h3xstream.findsecbugs.injection.BasicInjectionDetector.getInjectionPoint()` this is not handled properly. `invoke.getClassName(cpg)` results in the functional interface method name, in this case the method `accept`( of the interface `Consumer`). `Repository.lookupClass()` then of course throws a `ClassNotFoundException` and reports the method as missing class.

kohlschuetter:
When analyzing Java bytecode for "invoke" instructions, we assumed that there always was a valid class name specified when calling InvokeInstruction#getClassName.

Unfortunately, this is not true for "invokedynamic" calls, where BCEL repurposes this method to get the method name.

This leads to confusing warnings of the type "The following classes needed for analysis were missing: makeConcatWithConstants", etc.

Add code that checks for the INVOKEDYNAMIC subclass, skipping these instructions whenever necessary.

Fixes #332

### DIFF
--- a/findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/injection/AbstractTaintDetector.java
+++ b/findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/injection/AbstractTaintDetector.java
@@ -32,6 +32,7 @@ import edu.umd.cs.findbugs.classfile.MethodDescriptor;
 import java.util.Iterator;
 import org.apache.bcel.classfile.Method;
 import org.apache.bcel.generic.ConstantPoolGen;
+import org.apache.bcel.generic.INVOKEDYNAMIC;
 import org.apache.bcel.generic.Instruction;
 import org.apache.bcel.generic.InstructionHandle;
 import org.apache.bcel.generic.InvokeInstruction;
@@ -114,7 +115,7 @@ public abstract class AbstractTaintDetector implements Detector {
             Location location = i.next();
             InstructionHandle handle = location.getHandle();
             Instruction instruction = handle.getInstruction();
-            if (!(instruction instanceof InvokeInstruction)) {
+            if (!(instruction instanceof InvokeInstruction) || instruction instanceof INVOKEDYNAMIC) {
                 continue;
             }
             InvokeInstruction invoke = (InvokeInstruction) instruction;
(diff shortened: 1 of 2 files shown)


kohlschuetter:
Build failure around `CrlfLogInjectionDetectorTest.detectResponseSplittingKotlin:66` is unrelated (also happens without this change)

topolik:
LGTM.

ObjectDeserializationDetector fix is obvious.

The fix in analyzing sink locations is also fine. We don't expect `invokedynamic` calls to represent sinks, they call the bootstrap method. If a sink happens inside a custom lambda code then the compiler creates another method for the code. 

The sink is reached in `lambda$unknown$1` method 
`5: invokespecial #7                  // Method java/io/File."<init>":(Ljava/lang/String;)V` 

And not in `unknown` method
 `0: invokedynamic #5,  0              // InvokeDynamic #1:apply:()Ljava/util/function/Function;`


```java
    public static void unknown(String fileName) {
        Function<String, File> fn = f -> {
            return new File(f);
        } ;
        fn.apply(fileName);
    }
```


```
  public static void unknown(java.lang.String);
    descriptor: (Ljava/lang/String;)V
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
      stack=2, locals=2, args_size=1
         0: invokedynamic #5,  0              // InvokeDynamic #1:apply:()Ljava/util/function/Function;
         5: astore_1
         6: aload_1
         7: aload_0
         8: invokeinterface #4,  2            // InterfaceMethod java/util/function/Function.apply:(Ljava/lang/Object;)Ljava/lang/Object;
        13: pop
        14: return

  private static java.io.File lambda$unknown$1(java.lang.String);
    descriptor: (Ljava/lang/String;)Ljava/io/File;
    flags: (0x100a) ACC_PRIVATE, ACC_STATIC, ACC_SYNTHETIC
    Code:
      stack=3, locals=1, args_size=1
         0: new           #6                  // class java/io/File
         3: dup
         4: aload_0
         5: invokespecial #7                  // Method java/io/File."<init>":(Ljava/lang/String;)V
         8: areturn

  1: #27 REF_invokeStatic java/lang/invoke/LambdaMetafactory.metafactory:(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodHandle;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/CallSite;
    Method arguments:
      #28 (Ljava/lang/Object;)Ljava/lang/Object;
      #35 REF_invokeStatic a.lambda$unknown$1:(Ljava/lang/String;)Ljava/io/File;
      #36 (Ljava/lang/String;)Ljava/io/File;

```

walkerbrianpatrick:
Does this fix just remove the warning? My impression of the issue is that invokedynamic was preventing the scanner from finding bugs inside of lambda expressions.

PR #752, opened 2024-12-10
Remove array signature from classname

### LINKED ISSUE #751: Bytecode inspection does not properly check java classnames given an array
While investigating a fix for https://github.com/find-sec-bugs/find-sec-bugs/issues/332 (see https://github.com/find-sec-bugs/find-sec-bugs/pull/750), I noticed that we also sometimes assume that an array type is a simple classname, causing the corresponding class lookup to fail.

That failure is reported to Spotbugs, but apparently ignored internally.

Instead of ignoring this, we should try to extract the simple class type, and check this one instead.

kohlschuetter:
When looking up a class name, we sometimes mistake an array type for a simple classname, causing a subsequent class lookup to fail.

That failure is reported to Spotbugs, but apparently ignored internally.

Instead of ignoring this, we should try to extract the simple class type, and check this one instead.

Fixes #751

### DIFF
--- a/findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/common/ByteCode.java
+++ b/findsecbugs-plugin/src/main/java/com/h3xstream/findsecbugs/common/ByteCode.java
@@ -169,4 +169,22 @@ public class ByteCode {
         return null;
     }
 
+    /**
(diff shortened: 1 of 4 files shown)


h3xstream (approved):
Thanks for the bug fix

PR #758, opened 2025-04-16
docs - add OWASP 2021 top 10 references

JuditKnoll:
In the messages the 2013 OWASP Top 10 is the latest referenced OWASP list. Since then OWASP released two more top 10 lists: the 2017 and the 2021 version. (The 2025 list is still WIP.)
This PR refers the [2021 OWASP Top 10 list](https://owasp.org/Top10/) elements, based on the list of mapped CWEs at the list elements and the already CWEs refered in `messages.xml`.

One more small change: added CWE reference to POTENTIAL_XML_INJECTION bugpattern in both the `findbugs.xml` and `messages.xml` files.


h3xstream:
I'm not a big fan of OWASP Top 10 documentation. But I guess we need to update the old links.

I will comment on those that I think don't corresponding to the bug type.

h3xstream:
☝️  Note: I could be very critical on some resource when I think it as little information for developers to help them understand risk or the potential mitigation.
That said you can challenge these. I am going very quickly through most of these. :)

h3xstream:
Just give me a few days I will do the edits on the PR directly.

I need to finish the testing for the latest release.

